PR #9, opened 2022-01-02
Use GCP Project ID for Pub/Sub

wmil:
Use GCP Project ID for Pub/Sub as SDM (Device Access) Project ID doesn't work.

Hit the following error (time and Project ID masked for privacy) while I'm sure the OAuth token has `https://www.googleapis.com/auth/pubsub` granted.  After a few trial and errors I've found if I use the GCP Project ID for Pub/Sub it would work perfectly.

The pull request in its current form clearly could break existing configuration since it mandates a new field.  I'm happy to make additional changes if needed.

```
[01/01/2022, 00:00:00] [homebridge-google-nest-sdm] Plugin initialization failed, there was a failure with event subscription. Did you read the readme: https://github.com/potmat/homebridge-google-nest-sdm#where-do-the-config-values-come-from StatusError: Requested project not found or user does not have access to it (project=01234567-89ab-cdef-0123-456789abcdef). Make sure to specify the unique project identifier and not the Google Cloud Console display name.
    at MessageStream._onEnd (/usr/local/lib/node_modules/homebridge-google-nest-sdm/node_modules/@google-cloud/pubsub/src/message-stream.ts:301:20)
    at MessageStream._onStatus (/usr/local/lib/node_modules/homebridge-google-nest-sdm/node_modules/@google-cloud/pubsub/src/message-stream.ts:342:12)
    at ClientDuplexStreamImpl.<anonymous> (/usr/local/lib/node_modules/homebridge-google-nest-sdm/node_modules/@google-cloud/pubsub/src/message-stream.ts:198:38)
    at Object.onceWrapper (node:events:510:26)
    at ClientDuplexStreamImpl.emit (node:events:390:28)
    at Object.onReceiveStatus (/usr/local/lib/node_modules/homebridge-google-nest-sdm/node_modules/@grpc/grpc-js/src/client.ts:675:16)
    at Object.onReceiveStatus (/usr/local/lib/node_modules/homebridge-google-nest-sdm/node_modules/@grpc/grpc-js/src/client-interceptors.ts:424:48)
    at /usr/local/lib/node_modules/homebridge-google-nest-sdm/node_modules/@grpc/grpc-js/src/call-stream.ts:323:24
    at processTicksAndRejections (node:internal/process/task_queues:78:11) {
  code: 5,
  details: 'Requested project not found or user does not have access to it (project=1234567-89ab-cdef-0123-456789abcdef). Make sure to specify the unique project identifier and not the Google Cloud Console display name.',
  metadata: Metadata { internalRepr: Map(0) {}, options: {} }
}
```

### DIFF
--- a/config.schema.json
+++ b/config.schema.json
@@ -17,7 +17,12 @@
         "required": true
       },
       "projectId": {
-        "title": "Project ID",
+        "title": "SDM Project ID",
+        "type": "string",
+        "required": true
+      },
+      "gcpProjectId": {
+        "title": "GCP Project ID",
         "type": "string",
         "required": true
       },
--- a/src/Config.ts
+++ b/src/Config.ts
@@ -2,6 +2,7 @@ export type Config = {
     clientId: string,
     clientSecret: string,
     projectId: string,
+    gcpProjectId: string,
     refreshToken: string,
     subscriptionId: string,
     vEncoder?: string
--- a/src/Platform.ts
+++ b/src/Platform.ts
@@ -34,7 +34,7 @@ export class Platform implements DynamicPlatformPlugin {
 
         const options = config as unknown as Config;
 
-        if (!options || !options.projectId || !options.clientId || !options.clientSecret || !options.refreshToken || !options.subscriptionId) {
+        if (!options || !options.projectId || !options.clientId || !options.clientSecret || !options.refreshToken || !options.subscriptionId || !options.gcpProjectId) {
             log.error(`${config.platform} is not configured correctly. The configuration provided was: ${JSON.stringify(options)}`)
             return;
         }
--- a/src/sdm/Api.ts
+++ b/src/sdm/Api.ts
@@ -38,7 +38,7 @@ export class SmartDeviceManagement {
 
         try {
             this.pubSubClient = new pubsub.PubSub({
-                projectId: config.projectId,
+                projectId: config.gcpProjectId,
                 credentials: {
                     // @ts-ignore
                     type: 'authorized_user',


potmat:
@wmil: Thank you so much for contributing!

Do you have any idea why this is necessary?  That is, why the SDM project ID works fine for me but not for you?

wmil:
I don't have an answer to that but I can share my train of thought.

The work flow to create these credentials --
1. As the owner of the GCP project owner -- Create a GCP project (`my-gcp-project`) and establish an OAuth client inside this GCP project.
2. As the owner of the Nest devices owner -- Create an SDM project which is directly linked to the account's Nest devices -- the SDM Project ID seems to be some concept only in SDM world, see `sdm-prod` below.
3. As the owner of the Nest devices owner -- Grant access so that the GCP project / OAuth client will be able to access resources inside the SDM project.
4. As the owner of the GCP project owner -- Create a Pub/Sub Subscription (`projects/my-gcp-project/subscriptions/my-subscription-id`) inside the GCP project to track a Topic exported by SDM (`sdm-prod` could be the GCP Project ID for SDM, as seen in the Topic path `projects/sdm-prod/topics/enterprise-*`).

Here we can note --
- The SDM Project ID seems to be a concept only in the SDM world (another GCP project, `sdm-prod`), and the GCP infrastructure is unaware of it.
- Our interaction with GCP Pub/Sub from Homebridge is to access a Subscription `my-subscription-id` inside our GCP project `my-gcp-project`, and if we follow GCP Pub/Sub docs, this should be the one to use in establishing the connection.

If this logic is right, it's reasonable to use GCP Project ID for Pub/Sub.  Why the SDM Project ID worked for you remains a mystery to me...

potmat:
When doing this step: https://github.com/potmat/homebridge-google-nest-sdm#where-do-the-config-values-come-from

Which project-id did you put in the URL?

jkossis:
I have run into this issue as well. @potmat +1 for this PR making its way through!

jkossis:
Breaking both project ids out into separate properties is necessary. Once you supply the gcp project id as the config's `projectId`, while pub/sub initialization now works, calls to the actual sdm interface will 404 since they are using the same config property (`projectId`), which doesn't contain the sdm project id.

potmat:
@jkossis and @wmil which project ID are you using when you do this step: https://github.com/potmat/homebridge-google-nest-sdm#where-do-the-config-values-come-from?  That is, do you substitute the SDM project ID into the URL or the GCP project ID?

jkossis:
@potmat in the section where you specify to use the modified url, to include the extra pub/sub scope, the sdm project id is substituted in.

@wmil's [comment](https://github.com/potmat/homebridge-google-nest-sdm/pull/9#issuecomment-1004521560) highlights this as well.

potmat:
@jkossis : Ok fair enough, I don't really understand why but I will merge this soon.  I'm sorry for the delay but we are back under various lockdowns/school closures here in Ontario right now so.... I have been delayed.

wmil:
@potmat Thank you, 1.0.20 works.

wmil:
@potmat Following your latest README update, the reason that it worked for you but not for me became clear.

The Subscription ID you used was in the form of `projects/my-gcp-project/subscriptions/my-subscription-id`.
The one I used was just `my-subscription-id`.

The terminology for the longer string seems to be "Subscription Name" on GCP and hence the confusion.

potmat:
Yes too late I realized that subscriptionId was the wrong name for the property.  I'm not sure I can change it now without breaking everyone's installations.  I've changed the name in the config UI to just "Subscription" and added the example to clear up confusion.

I wonder if this means we no longer need the gcpProjectId?

PR #46, opened 2022-12-04
Added an optional config to use a switch for Eco Mode on thermostats.

hn-git:
Default to the current behavior of using a custom characteristic.

Custom characteristics are not visible through the default Home application.

Using a switch within the accessory provides an option for anyone who wants to keep all automations compatible with the Home app.

### DIFF
--- a/config.schema.json
+++ b/config.schema.json
@@ -40,6 +40,11 @@
         "title": "Video Encoder",
         "type": "string",
         "required": false
+      },
+      "isEcoSwitchEnabled": {
+        "title": "Create a Switch instead of a custom characteristic to indicate and control Eco Mode for each thermostat",
+        "type": "boolean",
+        "required": false
       }
     }
   }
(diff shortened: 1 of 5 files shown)


potmat:
@hn-git: It seems possible to create automations with custom characteristics using an app like Eve:

image

What happens if you do this? Does the Home app throw up errors or cause problems?

hn-git:
Yes, it's possible to create using other apps. When the scene with this action is viewed in the Home app it displays the accessory but not the characteristic being updated (eg. it just shows the thermostat being on).

I prefer to keep everything managed through the Home app without using scenes so it was useful to me. 
Feel free to close if you don't want to support maintaining/customizing this.

potmat:
@hn-git: My personal preference has been to avoid putting in "fake" accessories like switches/lights just to get around the limitations of the Home app.

PR #48, opened 2022-12-07
Fix: Fix fahrenheit accuracy and step size

hn-git:
A step size of 1 causes an incorrect conversion
when displaying fahrenheit and inability to step by 1F

Use the homekit default of 0.1 to provide sufficient precision to fix this.

The Home app temperature widget limits the min step size for the target temperature to 0.5C and 1F.


potmat:
@hn-git: If this is the case then wouldn't it be simpler to not bother setting minStep at all?

potmat:
@hn-git:  Yes if the default is already 0.1 then I would prefer we just don't set anything at all.

potmat:
@hn-git : Try version 1.1.9

Issue #97, opened 2023-09-04
Request failed with status code 500


fbmg:
Hi I get this error on different times ...

[...]

PR #109, opened 2023-10-14
Fix Axios Image Request Issue

DannyZolp:
Fixes https://github.com/potmat/homebridge-google-nest-sdm/issues/97

The issue had to deal with Axios trying to process the ArrayBuffer as a JSON string, which causes the 500 error to be thrown (like in issue #97). This was fixed by requesting the image as a base64 string and then processing it on the client.


potmat:
Thank you so much for the PR @DannyZolp.  However this PR seems to change many unrelated things, mostly related to code style (e.g. quotation, spacing, indenting, line wrapping etc.).  This makes it extremely difficult for me to determine the scope of the actual functional changes.

Are you able to modify this PR or create another one that only contains functional changes?  I understand you may have a desired code style or want consistency but I think that would best be handled in a different PR.

PR #112, opened 2023-10-19
Fix Axios Image Request Issue

DannyZolp:
The issue had to deal with Axios trying to process the ArrayBuffer as a JSON string, which causes the 500 error to be thrown (like in issue https://github.com/potmat/homebridge-google-nest-sdm/issues/97). This was fixed by requesting the image as a base64 string and then processing it on the client.


potmat (approved):

potmat:
Added changes directly to master due to code unrelated changes to nearby lines.

potmat:
Should be included in v1.1.20

Issue #16, opened 2022-01-28
Add a button to "autofill" Refresh Token


potmat:
The process of getting the refresh token is difficult and error prone, especially for people who are not computer savvy (e.g. using a "curl" command).  It should be possible to have a button in the config UI that will open the authentication screen and get the value.

krehman:
Can you not potentially use the Terminal in Homebridge UI?

Daxulan:
yes i'm blocked at this point , i have all my code but curl is a mystery for me and i try but don't work ..

krehman:
I did this process a while back so don't really recall off the top of my head, but anywhere in the instructions, e.g. [Use the API](https://developers.google.com/nest/device-access/use-the-api), it asked to use the `curl` command, I just opened up the **Terminal** from the _Homebridge-UI_ under the menu and inputted the command.

![image](https://user-images.githubusercontent.com/6234445/171038438-6da20f1f-82b8-4785-ae89-34e5ae4a805d.png)

litex2x:
I don't even think the refresh token works properly? I keep having to get a new one every week.

PR #131, opened 2023-12-29
Implement Custom UI to Facilitate Refresh Token Generation

jvartanian:
Using the [Homebridge Custom Plugin UI](https://github.com/homebridge/plugin-ui-utils) library, implements a custom config UI that makes it easier to generate refresh tokens.

The new UI requires that the user input a client ID, client secret, SDM project ID, & PubSub subscription. Once these values are present, the user can click the `Initiate Google Login` button, which will open the Google OAuth URL in a new tab.

Once the user completes the login flow, they can copy the resulting `google.com` link into the `Auth URL` text box. A refresh token will then be automatically generated and populated into the config.

Screenshot 2023-12-30 at 9 38 08 AM

Addresses #16


potmat:
Wow, this is amazing! I’m afraid I’m out of the country for a week or two so I’ll review it as soon as I get back. I really appreciate your contribution making no mistake.

src/homebridge-ui/public/index.html
```diff
@@ -0,0 +1,77 @@
...
+    </div>
+
+    <h5 style="margin-top: 40px">Config</h5>
+    <hr class="header-divider"/>
+</div>
+
+<script src="https://cdn.jsdelivr.net/npm/lodash@4.17.21/lodash.min.js"></script>
+<script>
+    // Create a script tag to load the JS onto the page
+    const script = document.createElement('script')
```
potmat, line 70: Why do this instead of simply referencing the script directly?  Something about the cache? Is reloading the page not sufficient?

potmat:
When I open the settings screen on an already configured system I just see this:

image

potmat (changes requested):

src/homebridge-ui/public/index.html
jvartanian: This ensures that the latest `nest-sdm-ui.js` is being used. If the script's `src` doesn't change between releases, the browser will sometimes use a cached version, even through page reloads.

This solution was inspired by the [Ring plugin](https://github.com/dgreif/ring/blob/main/packages/homebridge-ring/homebridge-ui/public/index.html)

jvartanian: The alternative is to have all of the code within the HTML file (e.g. [homebridge-meross](https://github.com/bwp91/homebridge-meross/blob/latest/lib/homebridge-ui/public/index.html)). I generally like separating the code from the UI, but if you prefer it all-in-one, I can definitely make that change.

jvartanian:
> When I open the settings screen on an already configured system I just see this:
> 
> image

I'm having a hard time reproducing this issue. I just tested the upgrade path on my mac's homebridge server, and my raspberry pi's, both displayed the custom UI successfully. 

Do you see an error toast, or a console log?

potmat:
Hi @jvartanian, I'm so sorry this hasn't been merged yet.  I've been extremely busy with work and life.  I haven't forgotten, I promise I'll get to it.

PR #206, opened 2025-12-01
bump for homebridge v2 compatibility

gtalusan:
I only have a Nest Thermostat to test, but this plugin loads up without warnings in Homebridge v2.  The deprecated HapNode-JS and Homebridge APIs didn't affect the rest of the code in this plugin.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "homebridge-google-nest-sdm",
-  "version": "1.1.22",
-  "lockfileVersion": 2,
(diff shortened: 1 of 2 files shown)


potmat:
Hi @gtalusan, thanks for the PR!  However, I'm not sure that simply changing package.json counts as "supporting" v2.  Is the purpose of this PR to show that the plugin will load in that case?

gtalusan:
Hi @potmat, the plugin loads and works in my testing as described above.  We may call it "compatibility" if that is suitable.

Issue #149, opened 2024-02-12
Nest Thermostat keeps resetting temperature


aneillans:
All setup, and mostly seems to work ok, however, when I adjust the set point via home bridge the setpoint resets back to what it was - after correctly updating the Nest (verified via the Nest app).

Have I missed something or is this a bug? Any way to enable further diagnostic logging ?

aneillans:
Config

{
    "clientId": "<redacted>go.apps.googleusercontent.com",
    "clientSecret": "<redacted>",
    "projectId": "319e152c-ca3f-4152-<redacted>",
    "refreshToken": "<redacted>",
    "subscriptionId": "projects/nesthomebridge-<redacted>/subscriptions/Nest",
    "fanDuration": 900,
    "platform": "homebridge-google-nest-sdm"
}

potmat:
Enable debug mode in Homebridge settings and post some logs from when you try to set the temperature.

aneillans:
[23/02/2024, 10:47:09] [homebridge-google-nest-sdm] Triggered SET TargetTemperature:18.5 Family Room Thermostat
[23/02/2024, 10:47:09] [homebridge-google-nest-sdm] Executing command sdm.devices.commands.ThermostatTemperatureSetpoint.SetHeat with parameters {"heatCelsius":18.5} Family Room Thermostat
[23/02/2024, 10:47:10] [homebridge-google-nest-sdm] Execution of command sdm.devices.commands.ThermostatTemperatureSetpoint.SetHeat returned undefined Family Room Thermostat

potmat:
Are there any errors? Returning undefined isn't necessarily a problem. Maybe go further forward/backward in the logs?

potmat:
I see it setting the heat to 18, then to 16... 16 seconds later.  I have no idea why this would happen, are you dragging the slider in the app or something?

lukebur06:
I am having the same issue.

potmat:
Perhaps something with event subscription?  When I adjust the temperature via the home app at my house I first see Homebridge update the temperature, then an event from Google saying the temperature has been set.  Here I see no event.  Is there any errors on Homebridge startup about failure with event subscription?

aneillans:
Getting no errors anywhere logged :( Anything I can do to crank up the verbosity on logging other than just setting a debug flag ? 

Also seeing this being logged I noticed this morning, not sure if its also related:

[...]

PR #207, opened 2025-12-14
Fix HeatCool range resetting to previous value

jharris1829:
Added additional logic when in HeatCool mode to deal with HomeKit sending both temperature values anytime there is a change, resulting in two SDM calls. The code now waits for both values before executing the change. I have tested on my own home and this fixes issue #149. This should also reduce the rate limiting problems since only one request is being sent instead of two for every change.


walking:
can it be merged?

PR #209, opened 2025-12-17
Add comprehensive setup guide with automated and manual installation scripts

WeekendSuperhero:
## Overview

This PR adds a comprehensive setup guide and automation scripts to streamline the Google Cloud Platform configuration process for homebridge-google-nest-sdm.

### 📄 Documentation
- **SETUP.md** - Complete setup guide with both automated and manual installation paths
  - Clear prerequisites and verification steps
  - Step-by-step instructions for both gcloud CLI and Google Cloud Console
  - Comprehensive troubleshooting section
  - Configuration reference with all available parameters

### 🤖 Automated Setup Script
- **scripts/setup-nest-sdm.sh** - Interactive automation script that:
  - Validates prerequisites (gcloud CLI, jq)
  - Creates or uses existing GCP projects
  - Enables required APIs (Smart Device Management, Pub/Sub)
  - Creates Pub/Sub topics and subscriptions with proper permissions
  - Guides through manual OAuth configuration steps
  - Generates refresh tokens
  - Tests API connectivity
  - Outputs ready-to-use Homebridge configuration JSON

### 🔍 Verification Script
- **scripts/verify-nest-sdm.sh** - Comprehensive troubleshooting tool that:
  - Validates all configuration settings
  - Tests API access and authentication
  - Verifies device discovery
  - Checks Pub/Sub configuration and permissions
  - Confirms enabled APIs
  - Supports interactive and automated modes
  - Generates access tokens for manual testing

## Key Features

- ✅ Support for both new and existing GCP projects
- ✅ Automated prerequisite checking
- ✅ Clear error messages and warnings
- ✅ Color-coded output for better readability
- ✅ Comprehensive troubleshooting guidance
- ✅ Verification tool to validate setup

## Benefits

1. **Simplified Setup**: Reduces setup time from 1-2 hours to ~15 minutes for automated path
2. **Error Prevention**: Validates configuration at each step
3. **Better Documentation**: Clear, organized guide covering all scenarios
4. **Troubleshooting Support**: Built-in verification tool to diagnose issues
5. **Flexibility**: Supports both automated and manual setup approaches

### SETUP.md
- Added comprehensive table of contents
- Documented automated and manual setup paths
- Included troubleshooting section with verification script usage
- Added configuration parameter reference table
- Fixed typos and improved clarity
- Clarified support for existing GCP projects

### Scripts
- Created `setup-nest-sdm.sh` - fully automated setup script
- Created `verify-nest-sdm.sh` - configuration validation tool
- Clean, well-structured code with proper error handling
- Interactive prompts with helpful context

## Testing

- ✅ Scripts tested with fresh GCP project creation
- ✅ Scripts tested with existing GCP projects
- ✅ Verification script validates all configuration aspects
- ✅ Documentation verified for accuracy

## Notes

- The $5 Device Access registration fee still requires manual payment
- OAuth consent screen and credentials require browser interaction
- Scripts handle both new and existing GCP project scenarios
- All manual steps are clearly documented with direct links to GCP console

This PR addresses the complexity of initial setup that many users struggle with, making it much easier for new users to get started with the plugin.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,3 +104,4 @@ typings/
 
 # TernJS port file
 .tern-port
+nest-sdm-credentials.json
--- a/README.md
+++ b/README.md
@@ -3,21 +3,51 @@
 
 # homebridge-google-nest-sdm
 
-A Homebridge plugin for Google Nest devices that uses the [Google Smart Device Management API](https://developers.google.com/nest/device-access). Supports Cameras, Doorbells, Displays, and Thermostats.  Supports HomeKit Secure Video (please read the section on HKSV below).
+A Homebridge plugin for Google Nest devices that uses the [Google Smart Device Management API](https://developers.google.com/nest/device-access). Supports Cameras, Doorbells, Displays, and Thermostats. Supports HomeKit Secure Video (please read the section on HKSV below).
 
 **Please read the [FAQ](https://github.com/potmat/homebridge-google-nest-sdm#faq) before creating an issue.** If you are having trouble with the setup process you can try reaching out to others in [Discussions](https://github.com/potmat/homebridge-google-nest-sdm/discussions), on [Discord](https://discord.gg/kqNCe2D), or [Reddit](https://www.reddit.com/r/homebridge/), some people there may be able to help.
 
-
-
 # Disclaimer
 
-This package is not affiliated with, provided, endorsed, or supported by Google in any way.  It is intended for personal, non-commercial use only.  Please review the [Google Smart Device Management Terms of Service](https://developers.google.com/nest/device-access/tos) to ensure that your usage of this package is not in violation.
+This package is not affiliated with, provided, endorsed, or supported by Google in any way. It is intended for personal, non-commercial use only. Please review the [Google Smart Device Management Terms of Service](https://developers.google.com/nest/device-access/tos) to ensure that your usage of this package is not in violation.
 
 # Installation
 
-``npm install -g --unsafe-perm homebridge-google-nest-sdm``
+`npm install -g --unsafe-perm homebridge-google-nest-sdm`
 
-Don't forget the ``--unsafe-perm`` part!
+Don't forget the `--unsafe-perm` part!
+
+# Easy Setup Guide
+
+**NEW!** We now provide an automated setup script that streamlines the Google Cloud Platform configuration process.
+
+### Quick Start
+
+For a simplified setup experience, check out our **[comprehensive setup guide](SETUP.md)** with automated scripts that handle most of the GCP configuration for you.
+
+**Features:**
+
+- 🤖 Automated GCP project creation and API enablement
+- ✅ Interactive guided setup process
+- 🔍 Built-in verification and troubleshooting tools
+- 📋 Generates ready-to-use Homebridge configuration
+- ⚡ Reduces setup time from 1-2 hours to ~15 minutes
+
+**From the repository root, run:**
+
+```bash
+./scripts/setup-nest-sdm.sh
+```
+
+The script will guide you through the entire setup process, including the manual OAuth steps, and generate your configuration file automatically.
+
+For verification and troubleshooting:
+
+```bash
+./scripts/verify-nest-sdm.sh nest-sdm-credentials.json
+```
+
+See the **[SETUP.md](SETUP.md)** guide for detailed instructions, or continue reading below for manual setup.
 
 # Example Homebridge config:
 
@@ -38,9 +68,9 @@ I recommend you use the plugin config UI to enter these values.
 
 # Where do the config values come from?
 
-Follow the getting started guide here: https://developers.google.com/nest/device-access/get-started  Please mind the "ONE IMPORTANT DIFFERENCE" section below.
+Follow the getting started guide here: https://developers.google.com/nest/device-access/get-started Please mind the "ONE IMPORTANT DIFFERENCE" section below.
 
-**clientId** and **clientSecret** come from this step: https://developers.google.com/nest/device-access/get-started#set_up_google_cloud_platform.  **clientId** should look something like "780816631155-gbvyo1o7r2pn95qc4ei9d61io4uh48hl.apps.googleusercontent.com". **clientSecret** will be a random string of letters, numbers, and dashes.
+**clientId** and **clientSecret** come from this step: https://developers.google.com/nest/device-access/get-started#set_up_google_cloud_platform. **clientId** should look something like "780816631155-gbvyo1o7r2pn95qc4ei9d61io4uh48hl.apps.googleusercontent.com". **clientSecret** will be a random string of letters, numbers, and dashes.
 
 **projectId** comes from this step: https://developers.google.com/nest/device-access/get-started#create_a_device_access_project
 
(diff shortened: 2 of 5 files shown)


ajplotkin:
This has sat since December without review, which seems a shame. I've set this plugin up end to end more than once, maintain a separate install guide for it, and today ran your script's flow against a genuinely clean Google account — new project, real $5 Device Access registration, real cameras. It got me most of the way, and the shape here is right.

**First, a correction to my own documentation that this PR prompted.** I had written that Google Device Access setup "cannot be scripted", lumping four things together: creating the Cloud project, the OAuth consent screen, the $5 registration, and getting a refresh token. Two of those four plainly *are* scriptable — `gcloud projects create` and `gcloud services enable` cover the project and APIs, and your Step 9 handles the token exchange. Only the $5 registration and the consent screen genuinely need a browser, and your script is honest about exactly that (`This step cannot be automated due to payment requirement`). I've fixed my guide, and adapted your `gcloud` sequence into it with credit — ISC → MIT, so the licensing is clean, but I'd rather say so out loud than have you find it. If you'd prefer I take it down or word the attribution differently, say so and I will.

That distinction matters more than it sounds. "You must click through two Google consoles" is a very different barrier from "you must click through all of it", and in my experience the bulk of setup pain on this repo isn't the clicking — it's the Pub/Sub topic/subscription/IAM plumbing, which is precisely what Steps 4–8 automate. Your sequence works: project, both APIs, topic, `sdm-publisher` binding and subscription all behaved exactly as written, and the two details people most often get wrong — self-hosted topic for post-Jan-2025 projects, and both scopes in the auth URL — you already handle.

## Four things a clean-account run hits

These only surface against a genuinely empty account, which is a hard thing to test for yourself. All four are "setup dies here with no useful error".

**1. Ordering: the Pub/Sub topic must exist before the Device Access project.** Step 6 creates the Device Access project and Step 7 creates the topic. But the Device Access Console demands a topic *the moment you tick "Enable events"* — required field, validated on the spot against `projects/{project}/topics/{topic}`, no way past it. So you arrive at a mandatory field for a topic that won't exist until the next step, with nothing telling you what to type. I hit this myself, because I'd adapted your ordering without questioning it. Swapping the two steps fixes it, and the registration step can then print the exact string to paste.

**2. Publishing status — the one I'd most want the script to prompt for.** While the consent screen is in **Testing** with **External** user type, Google revokes refresh tokens after **7 days**. The setup works perfectly and then silently stops a week later, surfacing as `invalid_grant`, which points nowhere useful. SDM uses a restricted scope so you'll never pass verification and don't need to — publishing just adds an "Advanced → Go to (unsafe)" click at authorization and makes the token permanent. Nobody discovers this until their cameras die the following week.

**3. Test users are a hard gate, not advice.** Without the authorizing account added under Test users, the flow ends at `Error 403: access_denied` — "can only be accessed by developer-approved testers". The instruction currently sits inside a "press Enter when done" block, which is easy to walk past, and the failure appears four steps later on a Google error page with no link back to the cause.

**4. Client secrets can no longer be re-read.** Google now shows only the last four characters and offers no download after creation. If the secret doesn't match, the token exchange fails with `invalid_client` — which reads like a code problem but isn't. The fix is to add a *second* secret on the client (it doesn't invalidate the first) and use the JSON offered at creation, since that's the only copy you get. That also argues for reading the secret from the downloaded `client_secret_*.json` rather than prompting for a paste — I switched to that after fumbling it myself, and it removes a whole class of error.

## Two smaller notes

**`verify-nest-sdm.sh` is the more broadly useful half, and it already stands alone.** It takes `[config-file]`, defaults to `nest-sdm-credentials.json`, and has its own `load_config`, so it runs against an existing install with no relationship to the setup path. That isn't obvious from SETUP.md's table of contents, where it reads as step 4 of an automated flow. Most failures I've seen aren't "setup did the wrong thing" but drift afterwards — a subscription deleted, IAM changed, a token revoked. A prominent "already set up and something broke? run this" entry point would get it used far more.

**One credential-leak path worth closing.** You clearly thought about this — line 335 truncates with `${REFRESH_TOKEN:0:20}...`. But the failure path just above it prints the entire token endpoint response:

```bash
if [ -z "$REFRESH_TOKEN" ]; then
    print_error "Failed to get refresh token!"
    echo "Response: $TOKEN_RESPONSE"
```

That branch fires when `refresh_token` is missing — exactly the common case where Google returns a valid `access_token` and no refresh token (re-authorising a client that already has a grant). So the path most likely to run is the one that prints a live credential to a terminal the user is about to paste into an issue. Truncating it, or printing only `.error` and `.error_description`, closes it. There's prior art here: #225 was specifically about not writing the OAuth token to the log.

## Offer

Happy to send a PR against your branch with the reordering and the two prompts if that's easier than patching it yourself.

@potmat — this is worth a look. Eight months open, it automates the step that generates the most setup traffic on this repo, and it now has someone who has run it start to finish on a clean account. One real defect (the token print) and nothing else I'd block on.

WeekendSuperhero:
> This has sat since December without review, which seems a shame. I've set this plugin up end to end more than once, maintain a separate install guide for it, and today ran your script's flow against a genuinely clean Google account — new project, real $5 Device Access registration, real cameras. It got me most of the way, and the shape here is right.
> 
> **First, a correction to my own documentation that this PR prompted.** I had written that Google Device Access setup "cannot be scripted", lumping four things together: creating the Cloud project, the OAuth consent screen, the $5 registration, and getting a refresh token. Two of those four plainly *are* scriptable — `gcloud projects create` and `gcloud services enable` cover the project and APIs, and your Step 9 handles the token exchange. Only the $5 registration and the consent screen genuinely need a browser, and your script is honest about exactly that (`This step cannot be automated due to payment requirement`). I've fixed my guide, and adapted your `gcloud` sequence into it with credit — ISC → MIT, so the licensing is clean, but I'd rather say so out loud than have you find it. If you'd prefer I take it down or word the attribution differently, say so and I will.
> 
> That distinction matters more than it sounds. "You must click through two Google consoles" is a very different barrier from "you must click through all of it", and in my experience the bulk of setup pain on this repo isn't the clicking — it's the Pub/Sub topic/subscription/IAM plumbing, which is precisely what Steps 4–8 automate. Your sequence works: project, both APIs, topic, `sdm-publisher` binding and subscription all behaved exactly as written, and the two details people most often get wrong — self-hosted topic for post-Jan-2025 projects, and both scopes in the auth URL — you already handle.
> 
> ## Four things a clean-account run hits
> 
> These only surface against a genuinely empty account, which is a hard thing to test for yourself. All four are "setup dies here with no useful error".
> 
> **1. Ordering: the Pub/Sub topic must exist before the Device Access project.** Step 6 creates the Device Access project and Step 7 creates the topic. But the Device Access Console demands a topic *the moment you tick "Enable events"* — required field, validated on the spot against `projects/{project}/topics/{topic}`, no way past it. So you arrive at a mandatory field for a topic that won't exist until the next step, with nothing telling you what to type. I hit this myself, because I'd adapted your ordering without questioning it. Swapping the two steps fixes it, and the registration step can then print the exact string to paste.
> 
> **2. Publishing status — the one I'd most want the script to prompt for.** While the consent screen is in **Testing** with **External** user type, Google revokes refresh tokens after **7 days**. The setup works perfectly and then silently stops a week later, surfacing as `invalid_grant`, which points nowhere useful. SDM uses a restricted scope so you'll never pass verification and don't need to — publishing just adds an "Advanced → Go to (unsafe)" click at authorization and makes the token permanent. Nobody discovers this until their cameras die the following week.
> 
> **3. Test users are a hard gate, not advice.** Without the authorizing account added under Test users, the flow ends at `Error 403: access_denied` — "can only be accessed by developer-approved testers". The instruction currently sits inside a "press Enter when done" block, which is easy to walk past, and the failure appears four steps later on a Google error page with no link back to the cause.
> 
> **4. Client secrets can no longer be re-read.** Google now shows only the last four characters and offers no download after creation. If the secret doesn't match, the token exchange fails with `invalid_client` — which reads like a code problem but isn't. The fix is to add a *second* secret on the client (it doesn't invalidate the first) and use the JSON offered at creation, since that's the only copy you get. That also argues for reading the secret from the downloaded `client_secret_*.json` rather than prompting for a paste — I switched to that after fumbling it myself, and it removes a whole class of error.
> 
> ## Two smaller notes
> 
> **`verify-nest-sdm.sh` is the more broadly useful half, and it already stands alone.** It takes `[config-file]`, defaults to `nest-sdm-credentials.json`, and has its own `load_config`, so it runs against an existing install with no relationship to the setup path. That isn't obvious from SETUP.md's table of contents, where it reads as step 4 of an automated flow. Most failures I've seen aren't "setup did the wrong thing" but drift afterwards — a subscription deleted, IAM changed, a token revoked. A prominent "already set up and something broke? run this" entry point would get it used far more.
> 
> **One credential-leak path worth closing.** You clearly thought about this — line 335 truncates with `${REFRESH_TOKEN:0:20}...`. But the failure path just above it prints the entire token endpoint response:
> 
> ```bash
> if [ -z "$REFRESH_TOKEN" ]; then
>     print_error "Failed to get refresh token!"
>     echo "Response: $TOKEN_RESPONSE"
> ```
> 
> That branch fires when `refresh_token` is missing — exactly the common case where Google returns a valid `access_token` and no refresh token (re-authorising a client that already has a grant). So the path most likely to run is the one that prints a live credential to a terminal the user is about to paste into an issue. Truncating it, or printing only `.error` and `.error_description`, closes it. There's prior art here: #225 was specifically about not writing the OAuth token to the log.
> 
> ## Offer
> 
> Happy to send a PR against your branch with the reordering and the two prompts if that's easier than patching it yourself.
> 
> @potmat — this is worth a look. Eight months open, it automates the step that generates the most setup traffic on this repo, and it now has someone who has run it start to finish on a clean account. One real defect (the token print) and nothing else I'd block on.
> 

@ajplotkin happy to fix the pr or have you create one against my pr. Just saw this.

PR #212, opened 2026-06-08
Speed up stream startup ~8s to ~2s; make analyzeduration/probesize configurable

littlepope81:
## Summary

Cuts camera stream startup (time from HomeKit's start request to video flowing) from **~8s to ~2s**, and exposes FFmpeg's `-analyzeduration` / `-probesize` as optional config fields. Every behavioral change was driven by per-stage timeline diagnostics that are included in the PR (debug-level `[startup +Nms]` / `[ffmpeg +Nms]` marks), so the log shows exactly where startup time goes on any camera.

## The three bottlenecks found and fixed

Measured on 5 SDM WebRTC cameras (Nest wired + battery cams, 1920x1088 and a 1600x1200 doorbell):

**1. FFmpeg waited for the camera's periodic SPS/PPS (worst case ~15s).**
The first keyframe is now requested via FIR (RFC 5104) alongside PLI. PLI alone gets answered with a bare IDR and no parameter sets, leaving FFmpeg to wait for the camera's periodic SPS; the FIR-elicited keyframe carries SPS/PPS in the same access unit, so FFmpeg has the stream dimensions at the first keyframe. (Verified with an RTP capture of the exact bytes handed to FFmpeg, and re-verified by the cache experiments below.)

**2. FFmpeg burned ~5s estimating the frame rate of a ~7fps stream.**
Even with dimensions in hand at the first keyframe, `avformat_find_stream_info` kept sampling frames for its fps estimate. `-fpsprobesize 0` skips only that sampling: probing returns as soon as codec parameters are known, while the full `analyzeduration` window remains available for cameras whose first keyframe arrives late. (An earlier attempt to just lower `analyzeduration` broke exactly such a camera — that knob caps how long FFmpeg will *wait*, which is why it's a config option here rather than a lowered default.) Video is stream-copied, so the unset fps metadata is never consumed.

**3. The sender trickled the first keyframe over 1.3–2.3s.**
An RTP capture showed the first IDR (~50–100 packets, ~60–90KB) paced at the sender's conservative initial bandwidth estimate (~375kbps). The plugin now sends REMB advertising 4Mbps as soon as the video track appears and repeats it with each keyframe request. The same IDR now arrives in ~100–250ms. (The very first attempt can land before DTLS is up; it fails as a caught debug-level message and the retry on first RTP succeeds.)

## What was tried and deliberately removed

An earlier iteration of this branch also cached each camera's SPS/PPS on disk and spliced them in-band before the first keyframe. Two live experiments showed it was redundant once FIR was in place:

- **Stale-cache test**: rotating the cached parameter sets across all 5 cameras (wrong resolution, wrong profile) changed nothing — the camera's own in-band SPS/PPS arrive with the FIR keyframe and supersede the injected ones before the IDR decodes.
- **Removal A/B**: deleting the cache entirely produced FFmpeg-side timings identical to primed runs (dimensions at +0.5–0.9s, probe complete at +1.0–1.1s after spawn).

So the cache, the synthetic RTP injection, and the sequence renumbering are **not** part of this PR — FIR gets the parameter sets there at the source.

## Measured result (same cameras, before → after)

| Stage | Before | After |
|---|---|---|
| First keyframe fully received | +3.5–4.0s | +1.5–1.8s |
| FFmpeg probe complete / output starts | +7.9s | +1.9–2.3s |

The remaining ~2s is the SDM `GenerateWebRtcStream` round-trip (~0.7–1.1s), ICE/DTLS (~0.4s), and the camera producing a keyframe (~0.3s).

## Config additions (purely additive)

`analyzeDuration` and `probeSize` are exposed as optional config fields with the plugin's long-standing values (15000000 / 100000000) as defaults, including schema descriptions warning that lowering them can break cameras whose parameter sets arrive late. Unset = unchanged behavior.

## Diagnostics (debug mode only)

- `[startup +Nms]` marks: offer/answer, ICE states, first RTP, first keyframe, keyframe fully received (with packet count), next-frame arrivals, first REMB.
- `[ffmpeg +Nms]` marks parsed from stderr: decoder Reinit, `Input #0`, `Output #0`, first `frame=`.
- `NEST_RTP_CAPTURE_DIR` env var: dumps the exact video RTP bytes handed to FFmpeg (length-prefixed) for offline analysis.


potmat:
Wow very interesting!

My first thought is that you can adjust video quality settings for the cameras.  If you load a cached state that is now incorrect what happens?

littlepope81:
Great question — I didn't want to answer it from theory, so I ran the experiment: I rotated the cached SPS/PPS across all 5 of my cameras (each primed with another camera's parameters — wrong resolution, and in some pairs wrong profile, High vs Baseline) and restarted.

**Result: every camera streamed normally at normal speed.** The debug timeline shows why: the FIR-elicited keyframe carries the camera's *own* SPS/PPS in the same access unit, right behind the injected stale ones, using the same parameter-set IDs — so the real ones supersede the stale ones before the IDR decodes (you can even see FFmpeg's decoder briefly reinit with the stale pix_fmt flags, then settle on the real ones). The cache also re-learns from every stream, so it self-heals after a single stream start.

But your question exposed something better: **if the camera's own parameter sets always arrive with the FIR keyframe, the cache isn't doing anything.** I tested that too — deleted the cache entirely and compared FFmpeg-side timings unprimed vs primed: identical (dimensions at +0.5–0.9s after spawn, probe complete at +1.0–1.1s, on all cameras).

So I've removed the whole mechanism from the PR — the cache file, the synthetic RTP injection, and the sequence renumbering (−500 lines, and it was the trickiest code in the diff). FIR gets the parameter sets there at the source; what remains is FIR+PLI, `-fpsprobesize 0`, the startup REMB, the diagnostics, and the config fields. Same measured ~2s startup. The PR description is updated to document both experiments.

src/NestStreamer.ts
```diff
@@ -42,9 +48,40 @@ export class RtspNestStreamer extends NestStreamer {
...
+            const file = path.join(captureDir, `nest-rtp-${name.replace(/[^a-zA-Z0-9]+/g, '_')}-${Date.now()}.rtpdump`);
+            try {
+                this.captureStream = fs.createWriteStream(file);
+                this.log.info(`Capturing video RTP sent to FFmpeg to ${file}`, name);
+            } catch (e: any) {
+                this.log.warn(`Could not open RTP capture file ${file}.`, e?.message ?? e);
+            }
+        }
+        // Write one forwarded packet to the capture file as [2-byte BE length][bytes].
+        const captureVideo = (buf: Buffer) => {
```
potmat: We have a lot of functions as variables, which makes readability difficult.  Can we move captureVideo, sendRemb, and requestKeyframe to class level functions of WebRtcNestStreamer?

ajplotkin:
Deployed this branch on real hardware — confirming the claimed improvement with measured timings.

**Setup:** Raspberry Pi 4 (arm64), 4x WEB_RTC-only Nest cameras (3 cams + 1 legacy Hello doorbell), `vEncoder: "copy"`, Homebridge 2.1.1, Node v24.18.0.

**Measured via the `[startup +Nms]` diagnostic marks (two sessions on the Upstairs Camera, consistent):**

| Mark | Run 1 | Run 2 |
|---|---|---|
| ICE completed | +85ms | +85ms |
| Offer sent | +86ms | +86ms |
| Answer from Nest | +994ms | +1143ms |
| Video track received | +1024ms | +1155ms |
| ICE connected | +1220ms | +1335ms |
| Connection connected | +1585ms | +1613ms |
| First video RTP | +1964ms | +2133ms |
| **REMB (4Mbps) sent** | +1971ms | +2137ms |
| **First keyframe (IDR)** | +1973ms | +2139ms |
| **First keyframe fully received** | **+2127ms (54 pkts)** | **+2292ms (47 pkts)** |

User-perceived tile-open in Apple Home: ~6s (the remaining ~4s is Apple's HAP/SRTP setup + Home app rendering -- not addressable from Homebridge).

With `vEncoder: "copy"`, the ffmpeg args include both `-fpsprobesize 0` (from this PR) and `-codec:v copy` -- they compose correctly, confirmed in the live command line.

One additional finding while testing: the Pub/Sub event handler in `Api.ts:68` has an unguarded `event.resourceUpdate.events` that crashes on `relationUpdate` events (no `resourceUpdate` field). Filed as #214.

littlepope81:
This is a great report — thank you for running it on real hardware. Independent confirmation on a Pi 4 (arm64) matters, because every number in the PR description came from x86 Docker; the mechanism holding across architectures and camera generations is exactly what I couldn't claim alone.

Your keyframe numbers are the ones I'd point at: **54 packets in 154ms** and **47 in 153ms**. Before the REMB change those same IDRs took 1.3–2.3s on my cameras, so you're reproducing the ~10x delivery improvement on completely different hardware. Also useful that your legacy Hello doorbell is in the mix — that's a camera generation I don't have.

Two notes on your results:

**The ~6s perceived vs ~2s measured gap.** Your read is right, and it's worth stating plainly for anyone else measuring: this PR only addresses the plugin→FFmpeg leg. What remains is Apple's HAP/SRTP setup plus Home app rendering, which no Homebridge plugin can touch. The `[startup +Nms]` and `[ffmpeg +Nms]` marks deliberately stop where our control stops.

**`-fpsprobesize 0` composing with `-codec:v copy`.** Thanks for verifying that on the live command line — that combination is the one I most wanted a second pair of eyes on, since it's the case where the snapshot/probe path is the only consumer of decoded video.

---

**On the `Api.ts:68` crash (#214) — confirmed, fixed, and it's worse than it looks.**

I reproduced it and opened #218 as a separate PR off `master`, so it can merge independently of this one.

The dereference is unguarded, and `ResourceRelationEvent` genuinely has no `resourceUpdate` field — but two details make it a process-level failure rather than a logged error:

1. The throw is inside the async `subscription.on('message', …)` callback, which runs long after the enclosing `try/catch` has returned. It surfaces as an **unhandled exception**.
2. `JSON.parse(message.data)` on the line above is unguarded too, so a malformed payload takes the same path. I only found that second one while building the repro harness.

I drove the real compiled handler with a stubbed Pub/Sub subscription. Before: `relationUpdate` throws `TypeError: Cannot read properties of undefined (reading 'events')` (your exact error) and malformed JSON throws `SyntaxError`. After: neither throws, and motion/trait dispatch is unchanged. The harness is in #218 if you want to run it against your own build.

One thing your report gave me that logs alone wouldn't: I checked 16 hours of my own logs and found **zero** `relationUpdate` events, so I had no way to hit this locally. It fires when a device is added, removed, or moved between structures/rooms — you presumably reorganized something in the Google Home app during setup. Worth knowing for anyone trying to reproduce it.

littlepope81:
Two follow-ups from continued testing on my hardware, both split out as standalone PRs off `master` so this one stays scoped to stream startup:

- **#218** — `Api.ts` crashes on `relationUpdate` events (the crash @ajplotkin filed as #214). Reproduced and fixed; also covers an unguarded `JSON.parse` on the same path.
- **#219** — Pub/Sub replays queued events on reconnect. After a restart I measured **210 motion events spanning four days** delivered in a three-second burst, firing phantom motion on four cameras and three HKSV recordings of nothing. `getEventImage()` already had a 30s freshness gate; the motion/ring trigger had none.

Neither depends on this PR. Worth noting #219 is only visible *because* of the diagnostics added here — the `[startup +Nms]` marks and per-event logging are what made a four-day backlog burst legible in the log.

potmat (changes requested):

src/NestStreamer.ts
littlepope81: Done in db0d252. `captureVideo`, `sendRemb`, and `requestKeyframe` are now private methods on `WebRtcNestStreamer`, with the state they were capturing promoted to typed instance fields (`videoTransceiver`, `videoTrack`, `firstRembSent`, `firSeq`) and `REMB_BITRATE` as a `static readonly`.

I also had to move `mark` — `sendRemb` calls it, so it couldn't stay a local. It reads `this.startTime` now, assigned where the `t0` local used to be. Shout if you'd rather that one stayed inline.

Pure refactor: identical RTCP construction and ordering, identical log/timing-mark strings, same guards, and the keyframe interval binds the method. The transceiver/track fields are set in `onTrack` before anything reads them.

potmat:
Hi @littlepope81 can you resolve conflicts?

littlepope81:
Conflicts resolved in 658748b — merged `master` in, PR now shows mergeable/clean.

Both conflicts were with changes of mine that landed while this sat open, so the resolution kept both sides:

- **`FfMpegProcess.ts`** — this branch had renamed the stderr chunk to `text` for the `[ffmpeg +Nms]` diagnostics; #220 added the blank-line skip and padding-probe suppression. Kept the `text` variable *and* both filters.
- **`NestStreamer.ts`** — #221's `!streamInfo` rate-limit guard versus this branch's `mark('received answer from Nest')`. Guard goes first: if `streamInfo` is undefined we did not actually receive an answer, so marking it would be wrong.

Rebuilt `dist/` and verified after the merge: 4 promoted private methods from the refactor you asked for, the padding-probe suppression, and both rate-limit guards all present and compiling clean.

potmat:
@littlepope81 : Oops, sorry, there's conflicts again, can you have a look?

littlepope81:
Conflicts resolved again in 35e4614 after #223 landed — mergeable/clean.

This one was benign: only the generated `.js.map` actually conflicted. `src/StreamingDelegate.ts` auto-merged, but since #223 touched `closeRecordingStream` and `RecordingSessionInfo` — the same areas — I verified the result rather than trusting a clean merge:

- #223 intact: `streamId` on `RecordingSessionInfo` (and set at assignment), the stale-close guard, both `Promise.resolve(...).catch(...)` teardown wrappers, and the `connectResolve?.()` settles in `HksvStreamer`.
- This branch intact: `-fpsprobesize 0`, the configurable `analyzeDuration`/`probeSize`, and the four private methods from the refactor you asked for.

`dist/` regenerated, `tsc` clean.

potmat (approved):

potmat:
@littlepope81 and @ajplotkin:  With all the recent PR merges can you get latest and confirm that everything is working before I create a release?  I'd like to handle the pub/sub one separately.

littlepope81:
@potmat Tested `master` at 350468f on real hardware — **looks good for a release from my side.**

Worth stating up front: I installed **master itself**, not my feature branch, so my unmerged work (snapshots, HKSV motion triggers, adaptive bitrate) was *not* present and could not mask anything. Cameras fell back to the placeholder tile as expected for master.

## Environment

x86 Docker (`homebridge/homebridge`), Homebridge 2.1.1, Node v24.18.0, 6 WEB_RTC cameras (4 cams + doorbell + a second doorbell), `vEncoder: "copy"`, debug mode on.

## Build / packaging

- `npm run build` from clean: 0 errors.
- **`dist/` matches a fresh build byte-for-byte** — no drift between committed artifacts and source, which given the number of PRs that landed felt worth checking explicitly.

## Regression checks for the merged fixes

Drove the real compiled handlers with the harnesses from each PR:

| Fix | Result |
|---|---|
| #218 relationUpdate crash | no uncaught throws; motion + trait events still dispatch; malformed JSON logged, not thrown |
| #219 stale event gate | fresh dispatched, 4-day-old dropped, 29s/31s boundary correct, unparseable timestamp fails open |
| #221 rate-limit guard | descriptive `Error`, not `TypeError: ...reading 'mediaSessionId'` |
| #220 log noise | 0 raw `Empty H.264 RTP packet` lines across the whole run |
| #212 + #223 | `-fpsprobesize 0`, configurable analyzeduration/probesize, stale-close guard, `connectResolve` settles — all present in the installed artifact |

## Live behavior

Restart: **10/10 accessories restored, 0 errors.**

Five camera sessions, all completed, REMB sent on 5/5:

| Session | REMB | Keyframe complete | Probe complete |
|---|---|---|---|
| 1 | 1323ms | 1397ms | **721ms** |
| 2 | 1345ms | 1555ms | 1113ms |
| 3 | 2069ms | 2272ms | 2049ms |
| 4 | 2460ms | 3225ms | 7499ms |
| 5 | 1269ms | 1343ms | **699ms** |

Session 4 is my slowest camera (lowest idle framerate) and is consistent with its historical behavior, not a regression. Whole run produced 680 log lines with zero plugin-level errors — for comparison, the same activity before #220 produced ~26,000.

## Scope and caveats

- **Excludes #216**, per your plan to handle pub/sub separately. Nothing here depends on it.
- All my cameras are WEB_RTC — the **RTSP path is untested by me**. #221 touches `RtspNestStreamer.initialize()` (guarding the stream response) so a legacy-camera confirmation would be worth having.
- x86 only. @ajplotkin has been running these on Pi 4 / arm64 and has the legacy Hello doorbell, so that's the natural other half of the coverage.

No blockers found. Happy to re-test anything specific before you tag.

potmat:
@ajplotkin : do you have a legacy camera? I'm away from home at the minute and cannot confirm.

ajplotkin:
Yes — a legacy Nest Hello (wired) doorbell plus 3 WEB_RTC cameras, all on a Pi 4 (arm64).

On confirming the merges: I've been running #212, #217, #218, #219 and #223 in production on that hardware, and they're solid across both camera generations — fast stream startup, HKSV recording through the new session-lifecycle code, the `relationUpdate` guard, and the replay/freshness gate (which I've watched fire both ways on live traffic: fresh events through, a genuine duplicate dropped), including on the legacy doorbell.

One honest caveat so you can weigh it: my setup layers a go2rtc RTSP path on top (snapshots + avoiding a second Google stream), so I'm exercising the merged code paths but not byte-for-byte vanilla master — a truly vanilla end-to-end I can't easily run without disrupting my live cameras. But the merged logic itself has held up here across both generations. Handling the pub/sub one (#216) separately sounds right. Also just opened #224 — a small follow-up to #223 that @littlepope81 flagged.

Issue #214, opened 2026-07-17
Pub/Sub handler crashes on relationUpdate events


ajplotkin:
Enabling Pub/Sub events for the first time crashes the plugin on the very first message Google sends.

**Stack trace:**
```
TypeError: Cannot read properties of undefined (reading 'events')
    at Subscription.<anonymous> (src/sdm/Api.ts:68:73)
    at Subscriber.<anonymous> (@google-cloud/pubsub/src/subscription.ts:329:38)
```

**Root cause:** `Api.ts:68` does:
```typescript
if (event.resourceUpdate.events) {
```
without guarding against events that have no `resourceUpdate` field. Google sends `relationUpdate` events (device/room relationship changes) when you first enable Pub/Sub — these have `relationUpdate` at the top level, not `resourceUpdate`. The `relationUpdate` handler was commented out in the source:
```typescript
// if ((event as Events.ResourceRelationEvent).relationUpdate) {
//     const resourceRelationtEvent = event as Events.ResourceRelationEvent;
// } else
if (event.resourceUpdate.events) {   // <-- throws here
```

**Fix (3 lines):**
```typescript
if (!event || !event.resourceUpdate) {
    log.debug('Event without resourceUpdate (e.g. relationUpdate), ignoring');
    return;
}
if (event.resourceUpdate.events) {
```

**Repro:** Enable Pub/Sub events on a Device Access project that has never had them. The first message Google publishes is a `relationUpdate` and it crashes the handler, killing the Homebridge process.

**Env:** homebridge-google-nest-sdm 1.1.23, Homebridge 2.1.1, Node 24.18.0

PR #216, opened 2026-07-19
Auto-reconnect the Pub/Sub event subscription (fixes events silently stopping)

ajplotkin:
## Problem

The Pub/Sub event subscription is created once in the constructor, and its failure handling can't recover:

```js
this.subscription = this.pubSubClient.subscription(config.subscriptionId);
this.subscription.on('message', ...);
this.subscription.on('error', error => {
    this.log.error(...);
    this.subscribed = false;   // logs, but never re-subscribes
});
// no 'close' handler at all
```

The streaming-pull connection behind `subscription.on('message')` is long-lived and **can drop silently** — a reboot, a network blip, a token hiccup, or a half-open stall. When it does, the plugin stops receiving **all** camera events permanently: no motion, person, or doorbell events reach HomeKit, and (for anyone using HomeKit Secure Video) recording is never triggered. It stays dead until Homebridge is manually restarted.

The drop is frequently **silent** — the connection half-opens and emits a `'close'` with no `'error'`, so nothing is even logged; events just stop. I hit this reproducibly after a host reboot: Google kept publishing (the backlog flushed the instant a fresh subscription connected), but the plugin's dead subscription never pulled anything.

## Fix

Move the subscription setup into a re-callable `subscribeToEvents()` that re-subscribes on `'error'` **and** `'close'`:

- **Exponential backoff** 5s → 60s, reset to 5s on healthy traffic.
- **Double-fire guard** — `'error'` and `'close'` often both fire for one drop; only one reconnect is scheduled.
- **Crash-safe teardown** — `close()` is called *before* `removeAllListeners()` (it flips `isOpen` synchronously), and its returned promise is caught: during an outage `close()` can reject on an ack-flush RPC failure, and on Node ≥ 15 an unhandled rejection **crashes the process** — turning "missed events" into "Homebridge down." A no-op `'error'` sink absorbs late errors forwarded during in-flight gRPC teardown.
- **Half-open backstop** — a periodic recycle recreates the subscription to catch stalls that emit neither `'error'` nor `'close'`. Pub/Sub re-delivers anything published during the brief gap, so no events are lost.

The message handler itself is unchanged — this PR is scoped to reconnection only.

## Testing

Running in production. After a real host reboot the log showed exactly the intended self-heal instead of a silent death:

```
Nest event subscription error; reconnecting in 5s   ChannelError: Failed to connect before the deadline
Nest event subscription established
```

and it has re-established through several subsequent disconnect cycles without intervention. The backoff/guard/crash-path state machine was also checked with a deterministic mock (error+close reconnects once; a rejecting `close()` produces no unhandled rejection; backoff doubles and resets on a healthy message).

Note: I verified behavior against the deployed build; I haven't run `tsc` locally on this branch, so if CI flags a typing nit I'll fix it promptly.


ajplotkin:
Cross-linking for sequencing: this PR relocates the Pub/Sub message handler into `subscribeToEvents()`, overlapping @littlepope81's #218 (crash guard, same handler) — I'll rebase this onto #218 once it merges so its guard isn't clobbered. And this PR's reconnect/recycle is the trigger for the replay that #219 gates, making #219 a companion to it. Ideal order: #218 + #219, then this on top.

littlepope81:
Read through this properly since we're in the same handler — it's careful work. The `close()`-before-`removeAllListeners()` ordering with the rejection absorbed, the late-error sink during gRPC teardown, backoff reset on healthy traffic, and the periodic recycle for half-open stalls are all things that are easy to miss and you didn't.

**One finding worth addressing: the reconnect path loses the constructor's exception protection.**

The first subscribe is inside the constructor's `try/catch`:

```ts
try {
    this.pubSubClient = new pubsub.PubSub({...});
    this.subscribeToEvents(config.subscriptionId);
} catch (error: any) { ... this.subscribed = false; }
```

but every subsequent one is called from a bare timer:

```ts
setTimeout(() => this.subscribeToEvents(subscriptionId), this.reconnectDelay);
```

and `subscribeToEvents()` opens with `this.pubSubClient!.subscription(subscriptionId)`. If that throws on a reconnect attempt — bad client state after an outage, an auth object that has gone sideways, anything synchronous — it throws inside a timer callback with no handler above it, which is an unhandled exception rather than a logged failure. The failure mode is the ugly one: the plugin dies on the reconnect that was supposed to save it.

Wrapping the body of `subscribeToEvents()` in a `try/catch` that logs and schedules another attempt (rather than rethrowing) would keep the self-healing property intact even when a reconnect attempt itself fails.

It's the same shape as #214, incidentally: a `try/catch` that looks like it covers the handler but doesn't, because the risky work happens later on a different tick. That's what made me look for it here.

Two smaller notes, both take-or-leave:

- The 12-hour recycle is a deliberate reconnect, so it will trigger a backlog redelivery twice a day. Harmless with #219's freshness gate in place, phantom motion twice a day without it — more evidence these belong together.
- If credentials are permanently bad, this retries every 60s forever with a `warn` each time. Self-healing is the right default and I wouldn't change the behavior, but capping the log (or backing off to a longer ceiling after N consecutive failures) would keep a misconfigured install from filling the log.

Happy to take the rebase in either direction, as I noted on #218.

potmat (changes requested):
No update to generated files? (dist directory)

ajplotkin:
Great catch — you're right, and it's the same shape as #214: the reconnect fires from a bare `setTimeout`, so a synchronous throw in `subscription()` (bad client/auth state after an outage) would be an unhandled exception on the very reconnect meant to self-heal.

Fixed in ba5b51f: `subscribeToEvents()`'s body is now wrapped in a `try/catch` that logs and reschedules with backoff instead of rethrowing, so a failed reconnect attempt just triggers the next attempt rather than crashing. Nice side effect — the *first* subscribe attempt now retries too, while a genuine client-creation failure still hits the constructor's catch.

Took your log-spam note as well: a shared consecutive-failure counter throttles both warn sites (first 5, then every 12th) and resets on healthy traffic, so a permanently-misconfigured install won't fill the log while keeping self-healing as the default.

Deployed and verified on our Pi 4 — subscription re-establishes cleanly, all cameras up, no crash. I left the handler guard out of this change so your #218 stays the single source of it on rebase. On sequencing: I'm happy either order and will defer to @potmat — and I'll gladly take you up on the rebase offer if #216 goes first.

ajplotkin:
Done — regenerated `dist/sdm/Api.js` + `.map` (`b550901`, pinned tsc 4.4.3). This branch also picked up the reconnect-path exception guard from @littlepope81's review (src `ba5b51f`), so the dist reflects both. Thanks!

potmat:
Hi @ajplotkin can you resolve conflicts?

ajplotkin:
Done — rebased onto `master`, conflicts resolved. Since #218 merged, I folded its `relationUpdate`/JSON handler guard *into* the new `subscribeToEvents()` (rather than reverting it), so both live together now.

While rebasing I also folded in a few things from review:
- the reconnect path's body is wrapped so a synchronous throw on a reconnect timer reschedules instead of crashing the process (the failure mode @littlepope81 flagged);
- the 12h recycle now uses a short fixed gap and doesn't ratchet the backoff (an idle subscription's recycle gap could otherwise grow past a downstream freshness window);
- a NOT_FOUND/PERMISSION_DENIED surfaces the readme hint so a misconfig stays diagnosable.

dist regenerated with the pinned tsc. Should be mergeable now.

littlepope81:
@potmat — on the generated-files question, I checked it so it isn't blocking on an unknown:

**`dist/` is in sync on this branch.** Checked out `f692b7d`, ran `npm run build`, and the committed `dist/` came back byte-identical — no drift.

**And it survives the merge.** Since the branch is based on a master that has moved 12 commits, I also merged `upstream/master` in and rebuilt, because a committed `dist/` is exactly where stale compiled output hides after a merge. Clean merge, no conflicts, and the rebuild changed **0 files** — so merging produces compiled output that genuinely matches the merged `src`. That concern looks resolved to me.

---

@ajplotkin — one interaction worth making explicit, since your recycle comment shows you were already thinking about it.

You correctly kept the deliberate recycle at a short fixed gap so it stays inside the event-freshness window:

```
RECYCLE_DELAY_MS = 2000       // 2s
MAX_EVENT_AGE_SECONDS = 30    // Camera.ts, from #219
```

2s vs 30s — safe, and the reasoning in your comment is right.

The failure path can't get the same protection, and I don't think it should. `RECONNECT_MAX_MS` is 60s, over the 30s gate — but the backoff isn't really the driver, the outage duration is. A five-minute network drop means the redelivered backlog is five minutes old no matter how promptly you reconnect, and #219 will discard it. That is what #219 is for: the bug it fixed was a 4-day backlog replaying as a wall of phantom motion.

So the honest characterization is: **reconnecting restores future events, it does not recover missed ones.** Anything older than 30s at redelivery is dropped by design. I think that's the right trade — a motion alert arriving minutes late is worse than no alert — but it's worth a line in the comment, because "auto-reconnect" reads like "no events lost," and someone will eventually file an issue asking where their events went during an outage.

Not asking for a code change on that. The `dist` verification above is the part that should unblock this.

ajplotkin:
Thanks for building it to check — the merge-and-rebuild especially, since a committed `dist` is exactly where stale output hides after a merge.

And you're right to state the reconnect semantics plainly: reconnecting restores *future* events, it doesn't recover ones missed during the outage. Anything past the 30s freshness gate (#219) at redelivery is dropped by design — which is the right trade (a motion alert minutes late is worse than none), but "auto-reconnect" does read like "no events lost," so being explicit helps. Happy to add that as a line in the reconnect comment if you'd want it in the code; otherwise this thread captures it.

littlepope81:
Yes — I'd put it in the code rather than leave it here.

This thread is findable by us, today. It is not findable by whoever is staring at the reconnect logic in six months wondering why a gap in their history didn't backfill after the subscription recovered. That person will read the comment, not the PR archaeology. The failure mode of leaving it out isn't a bug, it's someone re-deriving the interaction between reconnect and the freshness gate from scratch — or worse, "fixing" the gate because it looks like it's swallowing events.

Something like:

```ts
// Reconnecting restores the flow of FUTURE events; it does not recover events missed
// during the outage. Pub/Sub redelivers the backlog on resubscribe, but anything older
// than the freshness window (see #219) is discarded at that point — deliberately, since
// a motion alert delivered minutes late is worse than no alert. So an outage longer than
// that window means those events are gone, not merely delayed.
```

Wording is yours to take or leave; the part worth keeping is that it names *both* halves — reconnect restores future delivery, the gate drops stale backlog — because either one alone reads like a bug.

ajplotkin:
Added — `0aab909` on the branch, right above the `reconnect` closure so it's the first thing you hit reading that logic, and it names both halves (reconnect restores future delivery / the gate drops stale backlog) as you suggested. dist regenerated; comment-only, no logic change.

littlepope81:
Checked out `0aab909` and verified the claim rather than taking it:

- **Comment-only:** filtering the src diff to non-comment lines returns nothing — every added line is a comment.
- **dist in sync:** `npm run build` leaves the committed `dist/` untouched.
- **Emitted JS** changed by exactly those 5 lines in `dist/sdm/Api.js` plus the sourcemap, which is what a comment-only change looks like given comments are preserved in output here.

Placement is better than what I suggested, too — directly above the `reconnect` closure means you hit it before the logic rather than after, and pointing at `isEventStale` gives the reader somewhere to go. Nothing further from me on this one.

ajplotkin:
you're welcome. I am just happy to get all of this working.

ajplotkin:
@potmat this one's been sitting on a `CHANGES_REQUESTED` that I think was already answered, so flagging it rather than assuming you'd seen it.

You asked "No update to generated files? (dist directory)" on 21 Jul at 18:28Z. `f692b7d` went up at 19:38Z the same evening and includes `dist/sdm/Api.js` and its map alongside `src/sdm/Api.ts`; `0aab909` the next day does the same. So the branch has carried generated files since about an hour after you asked — the review state just never got dismissed.

Nothing else has changed on it. It's `MERGEABLE` against current master and littlepope81 independently reproduced the failure it fixes: his subscription was dead **13.5 hours** with no symptom until a container restart flushed 2547 backed-up events, ~1170 of which were correctly binned as stale. He's been running it four days without a manual restart since.

Happy to rebase if you'd rather have it fresh.

Issue #150, opened 2024-02-13
Last two updates has gobbled up memory


TanoopSungha:
After the last two updates I now find that it now gobbles up RAM on the Pi. When I restart the child bridge it then resolves the issue. I am using a Pi5.

potmat:
That's very odd, I can't account for it as there were no changes to the plugin itself, merely the arguments is passes to ffmpeg.  Can you see which process is using the RAM (Homebridge or ffmpeg)?

TanoopSungha:
The memory only goes down when I report or restart the child bridge to google nest sim

potmat:
Hi @TanoopSungha, I've not be able to reproduce this, my Pi4 still has 95% memory available after weeks of uptime.  I'll leave it open if anyone can reproduce it and submit a PR.

hillaker:
@potmat, I am seeing the same behavior as TanoopSungha on my Raspberry Pi 5-- memory slowly disappears over the course of the day until I have less than 0.5GB available (out of 8). 

Please find a screenshot of top sorted by memory-- it looks like ffmpeg is sticking around after a particular recording is complete. Happy to share anything else that might help identify a fix-- I'm running Kali linux, rather than Raspian or similar, and it's the ARM64 version. Not sure if that's relevant for this issue: 

![image](https://github.com/potmat/homebridge-google-nest-sdm/assets/5393987/f73de082-f457-4dbe-82b4-22e5601e8b45)

Memory after letting the plugin run for ~9 hours (after opening a stream, I left it running without opening a stream earlier in the day and didn't notice an increase in memory usage: 
<img width="411" alt="image" src="https://github.com/potmat/homebridge-google-nest-sdm/assets/5393987/c4dde5cd-cb9c-40d7-a803-7a797493a2b1">

Memory after restarting the plugin's bridge (but not home bridge), taken ~1 minute after the previous screenshot: 
<img width="404" alt="image" src="https://github.com/potmat/homebridge-google-nest-sdm/assets/5393987/7c2bf310-eae1-47d4-9b12-198ae1636629">

If there's something I can do to change my configuration to make it easier to diagnose the issue, please let me know and I'll give it a shot and share the results. Thanks!

Edited again to add that I only have this plugin as one that an outputs video-- my other plugins are switches.

GillesWi:
[...]

PR #217, opened 2026-07-20
Tear down the prior recording session before starting a new one (fixes orphaned ffmpeg)

ajplotkin:
### Problem

`handleRecordingStreamRequest` overwrites `this.recordingSessionInfo` without tearing down a prior session:

```ts
const hksvStreamer = new HksvStreamer(...);
this.recordingSessionInfo = { hksvStreamer, nestStreamer };  // clobbers any prior session
```

If a HomeKit hub starts a new recording before the previous one's close event fires (a brief hub reconnect, or rapid re-triggering under sustained motion), the previous session's `HksvStreamer` — and the ffmpeg child it spawned — is orphaned: nothing holds a reference to `destroy()` it, and the previous session's late close runs `closeRecordingStream`, which tears down the **new** `recordingSessionInfo` instead. The leaked ffmpeg keeps running, and under frequent motion these accumulate — growing memory over time. This is a plausible contributor to #150.

### Fix

Destroy any existing recording session before overwriting `recordingSessionInfo`, mirroring the `destroy()` + `teardown()` already done in `closeRecordingStream`.

### Notes

Found via adversarial review of the recording path. The change mirrors the existing teardown exactly (same two calls), so it's type-safe; I haven't run a full local build, happy to adjust if CI flags anything.


potmat:
No update to generated files?  (dist directory)

ajplotkin:
Good catch — done. Regenerated the compiled output (`4d6f669`): `dist/StreamingDelegate.js` + `.map`, built with the pinned tsc 4.4.3 so only the changed file moves. Thanks!

potmat (approved):

PR #218, opened 2026-07-21
fix: don't crash on relationUpdate events (#214)

### LINKED ISSUE #214: Pub/Sub handler crashes on relationUpdate events

littlepope81:
Fixes #214. Independent of #212 — branches off `master`, touches one file.

## The bug

`Api.ts` read `resourceUpdate.events` without checking that `resourceUpdate` exists:

```ts
if ((event as Events.ResourceEventEvent).resourceUpdate.events) {
```

Google also delivers **relation updates** (a device added to, removed from, or moved between structures/rooms). `ResourceRelationEvent` carries a `relationUpdate` and **no** `resourceUpdate`, so those events threw:

```
TypeError: Cannot read properties of undefined (reading 'events')
```

Two details that make it worse than a logged error:

- The throw happens inside the async `subscription.on('message', …)` callback, which runs long after the enclosing `try/catch` has returned — so it surfaced as an **unhandled exception**, not a caught one.
- `JSON.parse(message.data)` on the line above is unguarded too, so a malformed payload threw the same way.

The commented-out relation branch directly above the faulty line is what left the unguarded read as the leading condition. `Device.event()` already guards this exact shape (`(event as any).resourceUpdate && …`); `Api.ts` was the one place that didn't.

## The fix

Handle the relation case explicitly and bail, then dispatch events/traits off the verified `resourceUpdate`. Wrap the handler body so no single malformed message from an external source can take down the bridge. `message.ack()` still fires first, so nothing redelivers.

## Verification

Drove the **real compiled handler** (`dist/sdm/Api.js`) with a stubbed Pub/Sub subscription — no network, no credentials — across five payload shapes:

| Payload | Before | After |
|---|---|---|
| `relationUpdate` | **TypeError: Cannot read properties of undefined (reading 'events')** | no throw, debug-logged |
| malformed JSON | **SyntaxError (uncaught)** | no throw, error-logged |
| motion event | dispatched | dispatched |
| trait event | dispatched | dispatched |
| `resourceUpdate` with neither events nor traits | no throw | no throw |

<details>
<summary>Reproduction harness (standalone, no test framework needed)</summary>

```js
// node repro.js — stubs @google-cloud/pubsub so the constructor registers its
// callback against a fake subscription we can emit into.
const Module = require('module');
const REPO = process.cwd();
let captured = null;
const origLoad = Module._load;
Module._load = function (request) {
  if (request === 'googleapis') {
    return {Auth: {OAuth2Client: class { setCredentials() {} }},
            smartdevicemanagement_v1: {Smartdevicemanagement: class {}}};
  }
  if (request === '@google-cloud/pubsub') {
    return {PubSub: class {
      subscription() { return {on(evt, cb) { if (evt === 'message') captured = cb; }}; }
    }};
  }
  return origLoad.apply(this, arguments);
};
const {SmartDeviceManagement} = require(REPO + '/dist/sdm/Api.js');
const log = {debug(){}, info(){}, warn(){}, error(...a){ console.log('ERROR LOG:', ...a); }};
const sdm = new SmartDeviceManagement(
  {projectId:'p', clientId:'c', clientSecret:'s', refreshToken:'r', subscriptionId:'sub'}, log);
const DEVICE = 'enterprises/p/devices/DEVICE_A';
sdm.devices = [{getName: () => DEVICE, event: () => console.log('dispatched')}];

const relationUpdate = {eventId:'1', timestamp:'T', userId:'u',
  relationUpdate:{type:'UPDATED', subject:'enterprises/p/structures/S', object:DEVICE}};
try { captured({ack(){}, data: Buffer.from(JSON.stringify(relationUpdate))}); console.log('no throw'); }
catch (e) { console.log('THREW:', e.message); }
```
</details>

Built clean with `tsc`. Note the local Node 25 stub for `googleapis` in the harness is only because `jwa`→`buffer-equal-constant-time` fails to load on Node 25 (`SlowBuffer` removal) — unrelated to this change; the plugin itself runs on Node 24.


littlepope81:
Related sibling fix: #219 addresses a second problem in the same Pub/Sub ingestion path — Google replays a backlog of queued events on reconnect, and nothing gated the motion/ring trigger on event age. Measured on my hardware: after a restart, 210 motion events spanning four days of history arrived in a three-second burst, firing phantom HomeKit motion on four cameras and three HKSV recordings of nothing.

The two are independent and can merge in either order (#218 touches only `Api.ts`, #219 only `Camera.ts`/`Doorbell.ts`), though they're natural companions: this PR stops a malformed event from killing the process, #219 stops a stale one from firing a false alert.

ajplotkin:
Thanks for splitting this out — and for reproducing the crash we filed as #214. Confirming from the other side: we've run an equivalent guard in production (a `!event.resourceUpdate` bail + `try/catch` around the handler) since we hit that `relationUpdate` TypeError, so #218 matches what's worked for us.

Coordination note, since we're in the same code: our #216 (auto-reconnect) moves this handler into a `subscribeToEvents()` and, as written, carries an *unguarded* copy — so if #218 lands first, #216 would revert your guard. Happy to rebase #216 onto #218 once it merges so the guard is preserved.

Also worth flagging: #216's reconnect is exactly what makes #219 necessary — reconnecting triggers the backlog redelivery you measured, so without the freshness gate a live reconnect fires that phantom-motion burst. #218 + #219 + #216 really are one set. We've deployed #219's gate on our own boxes for that reason.

littlepope81:
Confirmed your coordination note — I checked #216's diff and it does carry the unguarded copy:

```ts
if ((event as Events.ResourceEventEvent).resourceUpdate.events) {
```

So merging #218 first would silently revert the guard. Two ways to sequence it, and I'm happy either way:

1. **#218 first, then you rebase #216** — as you offered. Straightforward, and #216's `subscribeToEvents()` just carries the guarded body forward.
2. **#216 first, then I rebase #218 onto it** — equally fine, and arguably simpler since #216 is the larger structural change. I'll do the work; no reason for the rebase burden to land on you when you're the one who found the original crash.

Whichever the maintainer prefers. If it helps, the guarded handler body is small enough to lift straight into `subscribeToEvents()` — it's the `!resourceUpdate` bail, the try/catch, and nothing else structural.

Your point that #216 makes #219 *necessary* rather than merely nice is the part I'd underline for review: a reconnect is precisely what triggers the backlog redelivery, so auto-reconnect without the freshness gate converts a silent-death bug into a recurring phantom-motion bug. #218 + #219 + #216 really are one set, and I don't think any of the three should land alone.

potmat (approved):

ajplotkin:
Thanks again for the generous rebase offer here. Happy either order — I'll defer to @potmat on which to review first. If #216 lands first I'll take you up on it; if #218 lands first I'll rebase #216 to carry your guard forward. (#216 now also has the reconnect-path exception fix you found, so `subscribeToEvents()` is a cleaner spot to fold the guard into either way.)

PR #219, opened 2026-07-21
fix: ignore stale replayed events instead of firing phantom motion

littlepope81:
Independent of #212 and #218 — branches off `master`, touches two files.

## The problem, observed in production

Google Pub/Sub redelivers queued events when the plugin reconnects. After a container restart I captured this in the logs:

- **210 motion events delivered in a 3-second burst**, immediately after startup
- payload timestamps spanned **four days** of history (oldest `2026-07-17T01:12Z`, delivered `2026-07-21T03:24Z`)
- they fired HomeKit motion on four cameras and started **3 HKSV recordings** of nothing

So every restart produces a wave of phantom motion alerts and junk recordings proportional to whatever Google has queued.

`getEventImage()` already refuses to fetch an image for an event older than 30s:

```ts
const dateDiff = (Date.now() - date.getTime())/1000;
if (dateDiff > 30) {
    this.log.debug(`Camera event image is too old (${dateDiff} sec), ignoring.`, this.getDisplayName());
    return;
}
```

…but nothing gated the motion/ring trigger itself, so the alerts fired anyway — and for WEB_RTC cameras the image path isn't even reached.

## The fix

- `Camera.isEventStale()`, checked in `Camera.event()` before dispatch.
- The same check in `Doorbell.event()`. It needs its own: `Doorbell` calls `super.event()`, and an early return inside `Camera.event()` does **not** stop the subclass from continuing its own `_.forEach`.
- The 30s window becomes one shared constant, reused by the existing image gate.
- **Unparseable timestamps fail open** (treated as fresh), so a malformed field can never silently swallow real motion.

## Verification

Drove the real compiled `Camera`/`Doorbell` handlers with a WEB_RTC trait stub (the path that reaches `onMotion`/`onRing` directly):

| Case | Before | After |
|---|---|---|
| motion, fresh (2s) | fires | fires |
| motion, **stale (4 days)** | **fires** | suppressed |
| motion, 29s (boundary) | fires | fires |
| motion, 31s (boundary) | **fires** | suppressed |
| doorbell ring, fresh (2s) | fires | fires |
| doorbell ring, **stale (4 days)** | **fires** | suppressed |
| unparseable timestamp | fires | fires (fail open) |

`npx tsc --noEmit` passes.

<details>
<summary>Reproduction harness (standalone, no test framework needed)</summary>

```js
const Module = require('module');
const REPO = process.cwd();
const origLoad = Module._load;
Module._load = function (request) {
  if (request === 'googleapis') {   // only needed on Node 25, unrelated to this change
    return {Auth: {OAuth2Client: class { setCredentials() {} }},
            smartdevicemanagement_v1: {Smartdevicemanagement: class {}}};
  }
  return origLoad.apply(this, arguments);
};
const {Camera} = require(REPO + '/dist/sdm/Camera.js');
const log = {debug(){}, info(){}, warn(){}, error(){}};
const device = () => ({
  name: 'enterprises/p/devices/A', parentRelations: [{displayName: 'Office'}],
  traits: {'sdm.devices.traits.CameraLiveStream': {supportedProtocols: ['WEB_RTC']}},
});
const mk = age => ({eventId:'e', userId:'u', eventThreadState:'STARTED',
  timestamp: new Date(Date.now() - age*1000).toISOString(),
  resourceUpdate:{name:'enterprises/p/devices/A',
    events:{'sdm.devices.events.CameraMotion.Motion':{eventSessionId:'s', eventId:'ev'}}}});

for (const [label, age] of [['fresh 2s', 2], ['stale 4 days', 4*86400]]) {
  const cam = new Camera({}, device(), log);
  let fired = 0; cam.onMotion = () => fired++;
  cam.event(mk(age));
  setTimeout(() => console.log(label, 'fired=', fired), 60);
}
```
</details>


ajplotkin:
Following up with independent production confirmation, since your #212 report valued the cross-hardware angle: I adopted the same `isEventStale` gate on our Pi 4 (4 WEB_RTC cameras + a legacy Hello doorbell — the generation you don't have), and it's now confirmed on both paths in the wild. Tonight's logs: three fresh motion events dispatched normally, and a genuinely stale event was dropped by the age check (the "event is too old" branch fired) — so the phantom-motion protection triggers on real traffic, not just the repro. Fresh events aren't false-dropped; stale ones are.

Thanks for splitting #218/#219 out cleanly, and for the detailed writeups — made adopting them straightforward.

littlepope81:
This is the verification I was missing — thank you. I'd been explicit in the PR that #219 was harness-verified only: I measured the *problem* in production (210 events across four days in a 3-second burst), but by the time the gate was deployed my Pub/Sub backlog had already drained, so I never watched the "too old" branch fire on live traffic.

You've now closed that gap on hardware I don't have, and on both paths — fresh events dispatching normally, a genuinely stale one dropped. That's the half I couldn't test.

Worth recording for whoever reviews this: between us the gate is now confirmed on x86 Docker + WEB_RTC cameras and on Pi 4 arm64 + a legacy Hello doorbell, with the boundary behavior (fresh through, stale dropped, unparseable timestamps failing open) covered by the harness in the PR description.

src/sdm/Camera.ts
```diff
@@ -114,8 +116,24 @@ export class Camera extends Device {
         }
     }
 
+    protected isEventStale(event: Events.Event): boolean {
+        const ageSeconds = (Date.now() - new Date(event.timestamp).getTime())/1000;
```
potmat, line 120: Do we need to guard against event.timestamp not parsing?  How much can we rely on event.timestamp being populated and parsable?  May not be an issue but....

potmat (changes requested):

src/sdm/Camera.ts
littlepope81: Good question, and the guard is already there — the terse comment is why it didn't read that way. `new Date(undefined)` and `new Date('nonsense')` both give `NaN` from `getTime()`, so a missing timestamp and an unparseable one land on the same branch:

```ts
if (Number.isNaN(ageSeconds)) return false;
```

It deliberately fails **open** — a bad timestamp lets the event through rather than dropping it, because the cost of being wrong that way is a duplicate alert, while the other way is a silently swallowed doorbell ring.

I've expanded the comment in 80ee89c to say all of that in place, so the next reader doesn't have to ask. The harness in the PR description already covers the case (`unparseable ts (fail open)` → fires).

On how much we can rely on the field: SDM always populates it in practice, but since it arrives as untrusted external JSON I'd rather not depend on that.

potmat (approved):

Issue #210, opened 2026-01-18
Nest Doorbell (wired) – WebRTC stream starts but ffmpeg reports Empty H.264 RTP packet, no live video in HomeKit


StefanZhang:
**Have you read the [FAQ](https://github.com/potmat/homebridge-google-nest-sdm#faq)?**

Yes

**Have you searched [existing issues](https://github.com/potmat/homebridge-google-nest-sdm/issues?q=is%3Aissue)?**

Yes

**Have you searched [existing discussions](https://github.com/potmat/homebridge-google-nest-sdm/discussions?discussions_q=)?**

Yes

**Describe the bug**

Hi, thanks for the plugin first of all.

I’m trying to use a **Google Nest Doorbell (wired)** with `homebridge-google-nest-sdm`.
Pairing, events, and doorbell notifications all work correctly, but **live video never appears in HomeKit** (Just that Google "G" logo).

### What works

* SDM authentication
* Device discovery
* Doorbell chime events
* `CameraLiveStream.GenerateWebRtcStream` command succeeds
* WebRTC ICE negotiation completes
* ffmpeg starts successfully

### What fails

Live video never appears.
ffmpeg continuously logs empty H.264 RTP payloads.

### Notes

* `mediaSessionId` is present and not undefined
* SDP looks valid and includes H.264 baseline profile
* RTP packets appear to arrive, but payload size seems to be zero
* Lowering resolution / FPS / bitrate does not change the behavior
* IPv4-only network also shows the same result

**Attach Logs**
```text
[homebridge-google-nest-sdm] Executing command sdm.devices.commands.CameraLiveStream.GenerateWebRtcStream
[homebridge-google-nest-sdm] Execution of command sdm.devices.commands.CameraLiveStream.GenerateWebRtcStream returned {
  "mediaSessionId": "***REDACTED***",
  "answerSdp": "v=0
  o=- 0 0 IN IP4 127.0.0.1
  s=-
  t=0 0
  m=video 9 UDP/TLS/RTP/SAVPF 97
  a=rtpmap:97 H264/90000
  a=fmtp:97 profile-level-id=42e01f;packetization-mode=1
  a=candidate:***REDACTED***"
}

[ffmpeg] [h264] Reinit context to 384x512
[ffmpeg] [sdp] Empty H.264 RTP packet
[ffmpeg] [sdp] Empty H.264 RTP packet
[ffmpeg] [sdp] Empty H.264 RTP packet
```
[...]

PR #220, opened 2026-07-21
fix: stop FFmpeg padding probes from drowning the debug log

littlepope81:
Independent of #212, #218, #219 — branches off `master`, touches one file.

## Measured problem

On a live bridge in **three minutes**: 26,400 log lines.

| Source | Lines | Share |
|---|---|---|
| `Empty H.264 RTP packet` | ~16,200 | 61% |
| blank continuation lines | ~9,000 | 34% |
| everything else (streams, events, diagnostics) | ~1,200 | 5% |

One camera alone (a doorbell) produced 8,090 of the blank lines. Debug logs are effectively unusable — the signal is 5% of the file.

## Two causes, both in `FfMpegProcess`

**1. Every FFmpeg message cost two log lines.** The stderr handler does:

```ts
data.toString().split(/\n/).forEach((line: string) => {
    log.debug(line, cameraName);
});
```

`split(/\n/)` yields a trailing empty string for every chunk, and it was logged too — rendering as a line containing just the camera name. That is the ~9,000.

**2. `Empty H.264 RTP packet` was logged once per packet.** It is normal WebRTC bandwidth probing (payload-less RTP), not an error, and it arrives constantly.

## Fix

- Skip lines whose trimmed value is empty.
- Count the padding probes rather than logging each one, and emit a single summary from the existing exit handler:
  `Suppressed 16183 'Empty H.264 RTP packet' messages (WebRTC padding probes).`

The summary keeps the signal: if the rate is ever abnormal, it is still visible in one line instead of sixteen thousand. Nothing else changes — the `started`/callback logic, error handler, and exit-code branching are untouched.

## Why it matters beyond tidiness

This is the exact message that led the reporter in #210 to conclude it was the cause of their no-video problem. It appears on healthy cameras constantly, so it is a standing source of misdiagnosis in bug reports.

`npx tsc --noEmit` passes.


littlepope81:
Deployed and measured on the live bridge. The effect is larger than the PR description estimated.

**Before** (3-minute window, six cameras, debug mode on): **26,400 lines** — ~16,200 padding probes plus ~9,000 blank continuation lines.

**After** (same bridge, same debug mode, six streams opened): **822 lines total**, of which:

- **0** raw `Empty H.264 RTP packet` lines
- **1** blank line in the entire file
- **5** summary lines, one per stream:

```
Suppressed 113 'Empty H.264 RTP packet' messages (WebRTC padding probes). Backyard camera
Suppressed  68 'Empty H.264 RTP packet' messages (WebRTC padding probes). Front Door
Suppressed 153 'Empty H.264 RTP packet' messages (WebRTC padding probes). Downstairs Camera
Suppressed 138 'Empty H.264 RTP packet' messages (WebRTC padding probes). Backyard camera
Suppressed  69 'Empty H.264 RTP packet' messages (WebRTC padding probes). Garage
```

Roughly a 30x reduction, and the per-stream counts still make an abnormal rate visible at a glance — which was the point of summarizing rather than dropping them outright.

Worth noting the debug log was genuinely unusable before this: real diagnostics were ~5% of the file, so any user asked to "send your debug log" was sending mostly padding probes.

src/FfMpegProcess.ts
```diff
@@ -59,6 +68,9 @@ export class FfmpegProcess {
             delegate.stopStream(sessionId);
         });
         this.process.on('exit', (code: number, signal: NodeJS.Signals) => {
+            if (this.emptyH264RtpPacketCount > 0) {
```
potmat: Do we really need to report how many were suppressed on exit?  I think just surpassing would be fine.

potmat (changes requested):

src/FfMpegProcess.ts
littlepope81: Fair — removed in 3ce108e. The counter and the exit-handler summary are gone; the padding probes are just dropped, and blank-line skipping is unchanged.

For the record, before/after on a live bridge was 26,400 lines in three minutes → 822, so the suppression alone carries essentially all of the benefit.

potmat (approved):

PR #223, opened 2026-07-21
Harden the HKSV recording lifecycle (hang on dead input, stale-close, teardown rejection)

ajplotkin:
Follow-up robustness to #217 (orphaned-ffmpeg session leak), from an adversarial review of the recording path:

**1. HKSV recording could hang forever if the input dies before ffmpeg connects.** `HksvStreamer` awaits `connectPromise` in `generator()`, but that promise is only ever *resolved* (in `handleConnection`) — never on failure. If ffmpeg exits/errors before connecting (a dead or misconfigured input), `handleDisconnect` fired but left `connectPromise` pending, so the generator suspended permanently: the recording hung, `handlingRecordingStreamingRequest` stayed true, and the listening server leaked. `destroy()` and `handleDisconnect()` now settle `connectPromise` (the generator then hits its existing `"Unexpected state!"` guard and fails fast) and `close()` the server.

**2. A late close of an orphaned session could kill the current recording.** `closeRecordingStream` destroyed `recordingSessionInfo` without checking which session the close was for. When a hub starts a new recording before the previous one's close fires (the exact sequence #217 addresses), the previous session's late `closeRecordingStream` tore down the **new** session. Now it compares `streamId` and ignores a stale close.

**3. Unhandled rejection from `nestStreamer.teardown()`.** It's async; a rejection was an unhandled rejection → bridge restart on Node ≥ 15. Guarded in both `closeRecordingStream` and the clobber-cleanup added in #217.

Found via a Fable code review while hardening a downstream deployment. `dist` regenerated with the pinned tsc; no behavior change on the happy path.


littlepope81:
Checked all three against the source rather than taking the description on faith — they hold up, and the first one is nastier than it reads.

**1. The `connectPromise` hang.** Confirmed. `connectResolve` is only ever called in `handleConnection`, so any path where ffmpeg dies before a client connects leaves it pending forever:

```ts
this.connectPromise = new Promise(resolve => this.connectResolve = resolve);
...
handleDisconnect() { this.socket?.destroy(); this.socket = undefined; }   // never settles it
destroy()         { this.childProcess?.kill(); ... }                      // never settles it
generator()       { await this.connectPromise; ... }                      // suspends forever
```

Worth emphasising that both `'error'` and `'exit'` route to `handleDisconnect`, so this catches the whole class, and the listening server leaked alongside it since `server.close()` also only happened on the success path. Resolving and letting the existing `"Unexpected state!"` guard fire is the right shape — fail fast and visibly beats a silent permanent hang.

**2. Stale close destroying the live session.** Confirmed — `closeRecordingStream` took `streamId` and ignored it entirely, so a late close for a replaced session tore down whichever session happened to be current.

One piece of supporting evidence you may not have: I have a change queued (`STOP_AFTER_MOTION_STOP = true`, so recordings end ~20-25s after motion stops instead of running until the controller hangs up) that makes this race **more** likely, because recordings end promptly and the next one starts sooner. So this guard becomes more load-bearing, not less. Flagging it since it argues for landing #223 regardless of what happens to my branch.

**3. Unhandled `teardown()` rejection.** Confirmed, and it's specifically the RTSP path that's exposed — `WebRtcNestStreamer.teardown()` wraps each step in its own try/catch, but `RtspNestStreamer.teardown()` is a bare `await this.camera.stopStream(...)`, so a rejection there propagates out of an un-awaited call. Node ≥ 15 turns that into a process exit.

---

**One optional hardening, take or leave.** When the generator now fails fast, the throw lands in `handleRecordingStreamRequest`'s `catch`, which logs but does not clear `handlingRecordingStreamingRequest` or `recordingSessionInfo` — those are only cleared in `closeRecordingStream`. In practice HAP should call `closeRecordingStream` after a generator error, so this is probably fine; I haven't traced hap-nodejs closely enough to claim otherwise. But a `finally` around that loop would make the reset independent of the controller behaving, which is the same "don't depend on the other side" reasoning behind fix 1.

Nice work — 1 and 3 are exactly the kind of thing that only shows up as "the bridge restarted and nobody knows why."

potmat (approved):

ajplotkin:
Thanks for the thorough pass — and good catch on the `finally`. You're right that #223's fail-fast made that path live: the generator now throws on a dead pre-connect input, and the loop's `catch` only logged, so `handlingRecordingStreamingRequest` would stick and refuse future recordings if HAP didn't call `closeRecordingStream`. Fixed in #224 — a `finally` calling `closeRecordingStream(streamId)` (idempotent, so a later HAP close is a no-op). Same "don't depend on the other side" reasoning as fix 1, as you said.

And thanks for the `STOP_AFTER_MOTION_STOP` context — good to know it makes the stale-close race more load-bearing, not less.

PR #224, opened 2026-07-21
Reset recording session in a finally (follow-up to #223)

ajplotkin:
Small follow-up to #223, from @littlepope81's review of it.

The recording generator's `for await` loop is wrapped in a `try/catch` that only logs on error — it never clears `handlingRecordingStreamingRequest`, which is reset only in `closeRecordingStream`. #223 made the `HksvStreamer` generator **fail fast** (throw) when the input dies before ffmpeg connects, so that error path now actually fires: if HAP doesn't then call `closeRecordingStream`, the flag stays `true` and every subsequent recording is refused.

This moves the reset into a `finally` that calls `closeRecordingStream(streamId)`, making cleanup independent of the controller's behaviour — the same "don't depend on the other side" reasoning behind #223's first fix. `closeRecordingStream` is idempotent (guarded on session identity and `hksvStreamer` presence), so a later HAP-driven close is a no-op.

Credit to @littlepope81 for flagging it on #223.


littlepope81:
Reviewed against master @ `350468f`. **The change is right and I'd merge it** — but I don't think the stated reason is the one that actually bites, and the `finally` is narrower than the problem it's guarding.

### 1. The flag is write-only on master

Every reference to it in `src/`:

```
src/StreamingDelegate.ts:87    protected handlingRecordingStreamingRequest = false;   // declaration
src/StreamingDelegate.ts:449   this.handlingRecordingStreamingRequest = false;        // write
src/StreamingDelegate.ts:482   this.handlingRecordingStreamingRequest = true;         // write
```

Nothing reads it — no subclass in `src/`, and it isn't a HAP-side field. So "every subsequent recording is refused" doesn't happen today; a stuck `true` is currently inert. Worth either wiring it up as a real guard or softening the claim, so the next reader isn't hunting for a refusal path that doesn't exist.

### 2. The real leak justifies the change anyway

The `catch` at 602 only logs. When the generator throws mid-stream, nothing releases:

- `hksvStreamer` → an orphaned **ffmpeg child process**
- `nestStreamer` → an **SDM stream still open on Google's side**

Those survive until the next recording hits the "tear down any prior session" block at 560–563, or indefinitely if no next recording comes. Given we're already hitting SDM rate limits (#221), holding a stream session open is the expensive half of this. That's the argument for this PR and it's a good one — it just isn't the flag.

### 3. The `finally` misses the setup phase — including #221's rate-limit throw

The `try` starts at 578, but the throwing awaits are above it:

```
545  const nestStreamer = await getStreamer(...);
546  const nestStream   = await nestStreamer.initialize();   // throws on SDM 429 (#221)
...
573  throw new Error('Streaming server already closed.')
...
578  try {
```

Both exit the generator without touching the new `finally`. The 546 path is the one I'd care about — that's exactly the rate-limit error #221 added, and it fired three times in my logs in one evening. Starting the `try` before 545 would make the cleanup guarantee genuinely unconditional, which is the stated goal.

### 4. Latent, not urgent

`handlingRecordingStreamingRequest = true` (482) lands before `recordingSessionInfo` is assigned (565), with two awaits in between. A previous session's `finally` landing in that window falls past the stale-session guard — `recordingSessionInfo` is still `undefined`, so the identity check doesn't fire — and clears the *new* session's flag. Inert today per (1), but it becomes real the moment the flag is wired up.

None of this blocks merging; (2) stands on its own. (3) is the one I'd fold in before it lands.

ajplotkin:
Thanks — this is a great review, and you're right on all four. Pushed a revision.

**1 — the flag is never read.** Correct. `handlingRecordingStreamingRequest` gets written but nothing reads it, so my "blocks future recordings" line (here and in my #223 reply) described something that can't happen — a stuck `true` does nothing today. Rewrote the comment to say what the change actually does and marked the flag unused. Wiring it up as a real guard is a behavior change I left for you to decide on; glad to do it as a follow-up.

**2 — the real problem is the leaked streams.** Agreed, and that's what the change is now about: on a throw, the ffmpeg child (`hksvStreamer`) and the open SDM stream (`nestStreamer`) never get released — and an open SDM stream matters more given the rate limits in #221.

**3 — the finally started too low.** Fixed. The `try` now starts at `getStreamer()`, so `initialize()`'s 429, a `getStreamer` failure, and the `'already closed'` throw all reach cleanup. Those happen before the session is stored in `recordingSessionInfo`, so the finally releases the local `nestStreamer`/`hksvStreamer` directly in that case.

Two things that came out of getting that right:
- Gating the finally on `recordingSessionInfo` identity is wrong — the normal HomeKit close calls `closeRecordingStream` first (clearing it) before the generator's finally runs, so identity-gating would send the normal path into the "release the locals" branch and call `nestStreamer.teardown()` a second time on every recording. Gated it on a `promoted` boolean instead, so teardown happens exactly once.
- On the 429 path, `initialize()` throws before the token is assigned, so I added a guard in `NestStreamer.teardown()` to skip `stopStream` when there's no token — otherwise the cleanup fires another doomed SDM call, which is the last thing you want while being rate-limited.

**4 — the pre-promotion flag window.** Agreed it's latent; left a note in the code. It only matters once the flag is wired up (see 1), and the `promoted` change already removes the case where a replaced session would clear the newer one's flag.

Diff is src + regenerated dist (pinned tsc 4.4.3), across `StreamingDelegate.ts` and `NestStreamer.ts`. Thanks again.

littlepope81:
Checked out `3a58f96` and verified rather than eyeballed. **This looks right to me — LGTM.**

**Build:** `npm run build` clean, and the committed `dist/` is byte-identical to a fresh build (no drift).

**The token guard, exercised against the real compiled `NestStreamer`:**

```
== RTSP: initialize() rate-limited (no token), then teardown ==
  PASS  initialize throws the #221 rate-limit error
  PASS  teardown makes NO stopStream call
== RTSP: normal path (token set) still stops the stream ==
  PASS  teardown calls stopStream exactly once
  PASS  stopStream got the real token
== WebRTC: teardown with no token (setup threw) ==
  PASS  teardown makes NO stopStream call
== WebRTC: teardown with a token set ==
  PASS  teardown calls stopStream exactly once
```

That was the failure mode I most wanted covered — piling another doomed `stopStream` onto an SDM API that just rate-limited us is exactly backwards.

**On the `promoted` boolean — your reasoning is right, and worth spelling out for whoever reads this next.** The subtle part is that it separates two different questions that both look like "is `recordingSessionInfo` set":

- *did we ever own these streamers* → `promoted`
- *do we still own them* → the identity check

Collapsing those into one identity test breaks precisely as you describe: after a normal close, `recordingSessionInfo` is already `undefined`, which reads as "never promoted" and drops into the release-the-locals branch, double-tearing-down every recording.

Worth noting the result is safe under **either** ordering of the HomeKit close vs. the generator's `finally`, which is what makes me comfortable with it:

- finally first → identity matches → `closeRecordingStream` tears down once; HAP's later call is the idempotent no-op
- HAP first → `recordingSessionInfo` cleared → finally skips → still exactly once

I walked the other exits too and didn't find a leak:

| exit | `promoted` | outcome |
|---|---|---|
| `getStreamer` throws | false | both locals undefined, flag reset |
| `initialize()` 429 | false | `nestStreamer.teardown()` → new guard short-circuits, no SDM call |
| `'already closed'` throw | true | now post-promotion → `closeRecordingStream`, once |
| generator throws mid-stream | true | identity matches → once |
| replaced by newer session | true | identity mismatch → skipped; the new session's prior-teardown block already released ours |

The flag not being reset in the skipped branches is also correct, for what it's worth: on a normal close HAP's `closeRecordingStream` already cleared it, and when a newer session replaced us it legitimately owns the `true`.

No further changes requested from me. Thanks for taking the review seriously and for the honest correction on the flag — that was my read too.

ajplotkin:
Appreciate you checking it out and running it against the real compiled streamer rather than taking my word for it — the token-guard cases are exactly the ones I was worried about. And thanks for spelling out the promoted-vs-identity distinction; that's a clearer statement of it than my comment. Thanks again for the whole review.

PR #226, opened 2026-07-23
Honor Google Home device renames (Info.customName)

littlepope81:
## What

Prefer the user-assigned device name (`Info.customName`) for the HomeKit accessory name, falling back to the current room-plus-type name.

## Why it matters

`getDisplayName()` derived the name only from `parentRelations` — which is the **room** the device is assigned to, not the device's own name. So when you rename a device in the Google Home app, the plugin keeps calling it by the room (e.g. `Office Camera`), and the only way to fix the HomeKit tile was to rename it again on the HomeKit side. Two places to keep in sync for one device.

After this, a rename in Google Home is the single source of truth: `customName` wins when set, and because it's read live from traits, an SDM `Info` event or the daily refresh picks it up **without a plugin restart**.

Resolution order:
1. `Info.customName` (trimmed) if set
2. room relation + device-type suffix (e.g. `Backyard Camera`) — unchanged behavior
3. `Unknown`

## Observed locally

Six cameras on a live bridge. Renamed five of them in Google Home (Front Door, Garage, Backyard, and two upstairs cameras); each HomeKit tile picked up the new name on the next event, no restart. Cameras with no custom name still show the room-derived name exactly as before.

## Notes

- Small refactor: the fallback expression was repeated in all four device classes, so it's now one `Device.resolveDisplayName(suffix)` helper.
- Verified the four resolution cases (custom set, custom blank, room-only, nothing known) against the compiled build; `tsc` clean, `dist/` regenerated.


littlepope81:
Added a follow-up commit: `getDisplayName()` reading `customName` only affects newly-created accessories, so a rename didn't reach an already-registered camera. The new commit updates the cached accessory's `displayName` and `Name` characteristic in `discoverDevices` when the device name changed — which is what actually propagates a Google Home rename to existing cameras.

PR #227, opened 2026-07-23
Advertise 2 concurrent streams, not one per resolution

littlepope81:
## What

Set `cameraStreamCount` to `2` instead of `camera.getResolutions().length`.

## Why it matters

`cameraStreamCount` tells HAP how many **concurrent** streams a camera supports — it creates one `RTPStreamManagement` service per count. It was being passed `getResolutions().length`, which is the length of the **resolution list** (11 entries), not a stream count. So every camera got **11 stream services**.

The visible effect: a wall of duplicate camera entries in the Homebridge accessories view (one per stream service), and an advertised concurrency far beyond what a Nest camera can serve. Two is the right number — a live view plus a HomeKit Secure Video hub recording can run at once, which is the real concurrent case.

HAP prunes the excess cached services on restore when the count shrinks, so existing installs clean themselves up on the next restart.

## Observed locally

Six cameras. Before: the accessories list showed each camera many times over (video/motion/etc. service duplicates stacked up), some marked "not connected." After dropping to 2 and restarting, each camera shows once, and live view + recording still work concurrently.

## Notes

`tsc` clean, `dist/` regenerated. One-line functional change.


ajplotkin:
The diagnosis here is right and worth merging — `cameraStreamCount: camera.getResolutions().length` really does create **eleven** `RTPStreamManagement` services per camera, and the two options are unrelated: `getResolutions()` is separately and correctly consumed as `streamingOptions.video.resolutions`, so the resolution list is unaffected either way. I hit this on my own install before finding your PR.

**A note on the value, and a correction to my own first version of this comment.**

I originally argued it should be `1`, on the grounds that hap-nodejs documents Secure Video cameras as exposing one stream and that "viewers go through the hub anyway". The first half is accurate; **the second half was wrong, and my own logs disprove it.** On a local network HomeKit clients stream *directly* from the accessory — here an iPhone at `192.168.1.42` receiving `srtp://192.168.1.42:51979` straight from Homebridge, while the home hub (`192.168.1.250`) appears only on the HDS connection carrying HKSV recording. The hub mediates recording, not local live view.

That matters for this PR, because it means `cameraStreamCount` is a real ceiling on **concurrent viewers**, not a formality. In a household with several phones, iPads and Macs, two people opening the same camera at once is not exotic — and with a low value the second one is simply refused.

So `2` is defensible and clearly better than 11. What I'd flag is only that the reasoning in the comment should be right: the second stream is **not** headroom for the recording. HKSV recording consumes no `RTPStreamManagement` service at all — it rides the HomeKit Data Stream and is wired via `recording.delegate`. So "one for live view plus one for the hub recording" describes something that isn't happening; the second service is purely a second *live viewer*. Same number, sounder justification, and it makes the case for going higher rather than lower if anyone reports contention.

For reference, hap-nodejs's own wording (`CameraController` options):

> Amount of parallel camera streams the accessory is capable of running. As of the official HAP specification non SecureVideo cameras have a minimum required amount of 2 (but 1 is also fine). Secure Video cameras just expose 1 stream. Default value: 1

I read "Secure Video cameras just expose 1" as describing Apple's own first-party behaviour rather than a constraint on what a bridge may advertise — and given live view is direct-to-device, a bridge serving a multi-device household has a real reason to advertise more.

Happy to send a commit to your branch if it's useful, but the change as written is fine by me — this is your find and your PR.

PR #229, opened 2026-07-23
Honor RECONFIGURE bitrate requests by re-advertising REMB

littlepope81:
## What

Act on the bitrate in HomeKit `RECONFIGURE` requests by re-advertising REMB, instead of ignoring them.

## Why it matters

When a viewer's available bandwidth changes (switching from wifi to cellular, say), HomeKit sends a `RECONFIGURE` with a new `max_bit_rate`. The delegate logged it and dropped it — the `(Ignored)` in the existing debug line.

Builds directly on the REMB work already merged (#212): the plugin already advertises a receiver-estimated max bitrate to pace the camera's sender. This makes that value adjustable at runtime. On a stream-copied feed resolution and fps genuinely can't change, but the bitrate ceiling can be re-advertised to the camera via REMB, which in copy mode adapts the HomeKit-facing rate directly.

- `NestStreamer.setMaxBitrate()` is a no-op by default (RTSP can't act on it).
- `WebRtcNestStreamer` overrides it: clamps to a 300kbps floor (so a pathologically low request can't starve keyframe delivery — HomeKit's lowest tiers sit around there anyway), updates the advertised value, and re-sends REMB.
- Idempotent: an unchanged bitrate is a no-op, so repeated RECONFIGUREs don't spam RTCP.

One race handled: a `RECONFIGURE` can arrive while `START` is still asynchronously setting up the session. Rather than drop it, the delegate stashes the bitrate and applies it when the session registers; the stash is cleared on stop.

## Observed locally

Six cameras across many live-view sessions. REMB re-advertises on reconfigure without disrupting the stream, and the first-REMB-at-startup behavior from #212 is unchanged. No RTCP spam from repeat requests (idempotency holds).

## Testing

Verified against the compiled `WebRtcNestStreamer`: below-floor clamp to 300k, idempotent no-op on an unchanged value, re-advertise on a real change, and no throw when a RECONFIGURE lands before the media track is set up. `tsc` clean, `dist/` regenerated.


potmat:
@littlepope81: Don't take this the wrong way but is there a person there or am I talking to Claude?

littlepope81:
A little of both!-pdOn Jul 28, 2026, at 08:15, Ted Potma ***@***.***> wrote:﻿potmat left a comment (potmat/homebridge-google-nest-sdm#229)
@littlepope81: Don't take this the wrong way but is there a person there or am I talking to Claude?

PR #230, opened 2026-07-23
Serve real snapshot tiles and advertise HKSV motion triggers

littlepope81:
## What

Two coupled additions to the camera/recording path:
1. **Real snapshot tiles** — HomeKit camera tiles show an actual recent frame instead of the placeholder logo.
2. **HKSV motion triggers** — cameras advertise motion as a recording trigger, and recordings end when motion stops.

They share the same FFmpeg/recording plumbing, so they're one PR.

## Why it matters

**Snapshots.** SDM has no snapshot API, so the plugin could only ever show a static placeholder (the Google "G" / Nest logo) on every camera tile. This adds a low-rate mjpeg output to the live and HKSV FFmpeg commands that keeps one JPEG per camera fresh on disk, served as the HomeKit snapshot:
- Writes are atomic (temp-file + rename), so a concurrent reader — or the other stream writing the same file — never sees a torn frame.
- `handleSnapshotRequest` prefers a cached **event** image on an EVENT-reason request (using the `ResourceRequestReason` enum), else serves the on-disk JPEG when it's fresh (< 24h) and structurally complete (validates the JPEG SOI…EOI markers), else falls back to the placeholder.
- If the snapshot directory is unavailable it degrades to the placeholder rather than breaking the whole FFmpeg command (and the stream).

**HKSV motion triggers.** Handing the accessory's `MotionSensor` to the camera controller makes HAP advertise `EventTriggerOption.MOTION` and link the sensor to `RecordingManagement`. Without it, plain (non-doorbell) cameras advertise an empty trigger set and motion recordings depend on Apple-hub heuristics. Recordings now end when motion stops instead of running until the controller closes them — on a camera with sparse motion the controller-driven path could otherwise run for hours, keeping the camera streaming continuously. This is behind `endRecordingOnMotionStop` (default `true`); set it `false` for the old run-until-closed behavior. The motion sensor decays 20s after the last event via a single clear-and-rearm timer that always fires, so it can't stick `true` and leave a recording running. A doorbell ring holds the same motion window open so ring-triggered recordings get the post-event tail.

## Observed locally

Six cameras on a live bridge. Tiles show real, recent frames that update (verified the on-disk JPEGs are valid images, ~20-55KB, one per camera, refreshing while streams run) instead of the placeholder. HKSV recordings fire on motion and end shortly after motion stops — bounded to tens of seconds instead of the multi-hour sessions seen before (one doorbell recording had run 3h50m under the old controller-driven path). No torn snapshots under concurrent live-view + recording.

## Notes / testing

- New config key `endRecordingOnMotionStop` (default true).
- `tsc` clean, `dist/` regenerated. Existing event/rate-limit behavior unchanged (verified against regression harnesses). Snapshot directory is created `0700` since the frames are interior views of the user's home.


ajplotkin:
I've been running a setup that solves the same two problems, so rather than +1 I thought the useful thing would be what's *additive* to what you've already built here.

First: **your snapshot approach is better than mine for the general case.** Piggybacking a low-rate mjpeg output onto the FFmpeg command that's already running gets a real tile with no extra infrastructure — mine keeps a warm stream in [go2rtc](https://github.com/AlexxIT/go2rtc) and reads a cached JPEG off tmpfs, which works well but means running another service just to have something to grab a frame from. The atomic temp-file+rename, the SOI/EOI structural check and degrading to the placeholder rather than breaking the stream are all details I'd have got wrong first time.

Everything below is measured on a Raspberry Pi 4 (arm64, 2 GB), four Nest cameras including a doorbell, plugin 1.1.24 / Homebridge 2.2.1. Written up with config and reasoning at **https://github.com/ajplotkin/nest-homekit-cameras**.

## Three things that bear directly on this PR

**1. Every HKSV clip currently records silent.** This matters here specifically, because advertising motion triggers will produce a lot more recordings. `videoArgs` starts with an unconditional `-an`, and `HksvStreamer` appends videoArgs *after* audioArgs, so it overrides the entire AAC-ELD block — the audio settings never take effect. [#234](https://github.com/potmat/homebridge-google-nest-sdm/issues/234), one-line fix in [#237](https://github.com/potmat/homebridge-google-nest-sdm/pull/237). Independent of everything else.

**2. "Recording ends when motion stops" runs into a hub behaviour that is invisible from the plugin side.** If a camera's Recording Options are People/Animals/Vehicles rather than Any Motion, the hub runs its own analysis and **silently discards a clip containing no qualifying subject** — `closeRecordingStream` returns `NORMAL`, nothing errors anywhere, and the clip simply never appears in the timeline.

That interacts badly with timing. Recording only starts once HomeKit asks, which is downstream of Pub/Sub: measured here, ~2 s for delivery plus ~1 s for FFmpeg to connect and reach a keyframe. And Google's own event timestamp lags its footage — for one doorbell press, Google's clip in the Home app began **6.1 s before** the timestamp it published to us. So the first recorded frame can be most of ten seconds after the person arrived, by which point they're out of frame and the hub bins the clip. That was far and away the biggest source of "it didn't record them" here, and no log on the plugin side shows it.

**3. The transcode isn't necessary**, which roughly halves the CPU cost of the extra recordings this PR will generate. Nothing hub-side validates delivered media against the negotiation — the strongest evidence is in this repo: the recording path has never applied a scale filter, so the plugin has been sending HomeKit un-negotiated resolutions for years without complaint. Copying the camera's H.264 removes the encoder entirely (~1.8 of 4 cores → remux only). [#235](https://github.com/potmat/homebridge-google-nest-sdm/issues/235) / [#238](https://github.com/potmat/homebridge-google-nest-sdm/pull/238). RTSP-protocol cameras still need to transcode, since the FIR/PLI keyframe loop only exists in `WebRtcNestStreamer`.

## The prebuffer — mentioning it, but not proposing it for this PR

The fix for problem 2 is to have footage from *before* the trigger. I keep a rolling in-memory ring per camera (remuxed `-c copy`, keyframe-aligned fragments, ~150 KB/s and a few percent of one core) and feed the recorder `[history][live]` when a trigger fires. On a real doorbell press yesterday: motion at `10:30:37`, recording requested the same second, **9.1 s of pre-trigger history served**, hub accepted the clip. It also implements the `prebufferLength: 4000` the plugin already advertises to HomeKit with nothing behind it ([#233](https://github.com/potmat/homebridge-google-nest-sdm/issues/233)).

But it needs a *continuously running* source, which stock doesn't have — the plugin dials per recording, so there's nothing to buffer. That means a hard go2rtc dependency, and I wouldn't suggest saddling this PR with one. Flagging it as the shape of a solution rather than proposing it here.

## On SDM quota, since it's the first objection to go2rtc

It reduces API pressure rather than adding to it: go2rtc holds **one** SDM session per camera and fans it out, so snapshots, live view and HKSV recording share a single session, where stock opens a dial for each. Measured here: one session per live camera, **0 direct SDM dials by the plugin in 24 h**.

The honest counterweight, from my own logs: I keep prebuffer rings for two cameras that are switched off, and those retry every five minutes and get `wrong status: 400` each time — about 107/hour across the two. That's a wart in my configuration rather than anything inherent, but it's not zero and I'd rather quote it than let someone discover it.

## Offer

Happy to test this against four cameras including a doorbell — I have the HKSV path fairly well instrumented at this point and could give you real numbers on the motion-trigger behaviour. And @potmat, the prebuffer is written as a self-contained module, opt-in and completely inert unless configured; happy to open it as a PR if you'd want it, or to leave it in my repo if a go2rtc dependency isn't something you'd want in the plugin. No expectation either way — #237 and #238 stand on their own regardless.

littlepope81:
Really useful comment — thanks. I'm traveling, so this is short and I'm testing remotely over the network; numbers will follow rather than lead.

**#237 and #238: adopted both, and running on my box.** Neither applied cleanly (I'd refactored `createRecordingSession` to thread a local `configuration`), so I re-applied by hand and moved `videoArgs` into a helper built after `getStreamer()`, since the copy decision depends on the streamer instance. My encoder tuning now only survives in the RTSP branch, which is correct — on WebRTC there's no encoder left to tune.

Deployed against six Nest cameras. The recording command it now builds, on a real WebRTC doorbell:

```
-i - -acodec libfdk_aac -profile:a aac_eld -ar 48k -b:a 64k -ac 1
     -f mp4 -sn -dn -codec:v copy
     -movflags frag_keyframe+empty_moov+default_base_moof tcp://127.0.0.1:33635
     -an -sn -dn -codec:v mjpeg -q:v 4 -vf fps=1/2,scale=640:-2 ... .jpg
```

AAC-ELD present with no `-an` ahead of it, `copy` on the mp4 output, and the only `-an` on the snapshot output where it belongs. Both fixes behave as you described.

Being straight about the limits: HomeKit didn't request a recording on either motion event, so I have the command but **not a saved clip yet** — no audible-playback confirmation and no fragment-cadence numbers from my own cameras. The cadence is the one I'd want measured before this is called settled, since `copy` drops the `-force_key_frames` guarantee and my cameras aren't going through go2rtc like yours.

One caveat on the CPU number that's now visible in that command: on a branch that also has this PR's snapshot output, `copy` saves less than you measured. The mjpeg output carries `-vf fps=1/2,scale=640:-2`, and that filter forces a decode. You drop the H.264 re-encode but keep the decode. Your setup doesn't hit this because your snapshots come off go2rtc rather than riding the recording command.

**On the prebuffer — yours is the right design, and mine isn't the same feature.** I read `PrebufferManager.js` before saying that. Mine is triggered from the motion handler, so the ring only starts filling once the event has been *delivered*; by construction it can't hold a frame from before the event. It hides the dial and keyframe time and nothing else. So it doesn't answer #233, and I've renamed it `motionPrewarm` to stop it claiming otherwise. Open yours.

Two things in your README I'd have got wrong: anchoring to the latched event timestamp rather than "now", and sizing from the hub's *selected* `prebufferLength` rather than the advertised max.

Where they're complementary is narrower than "both are prebuffers": on one event they don't stack, since a live ring means there's no dial left to pre-warm. But when the ring is configured and starved or empty for a camera — your 416 stall-kills, or a camera just added — the fallback today is a cold dial, and the prewarm makes it a warm one. And for anyone who never sets up go2rtc, it's the only thing helping. Ring first, prewarm as fallback, cold dial last.

**Yes please on the testing offer.** The case I'd most like measured is that tier boundary: what a recording actually does when the ring is configured but has no history yet for that camera. Hard to produce on demand, and it's exactly where the fallback either earns its place or doesn't.

@potmat — one correction while I'm here: `prebufferLength` can't simply be removed. It's mandatory in `CameraRecordingOptions` with a documented 4000 ms floor, so every plugin advertises at least that whether or not anything backs it. #233's only real fix is implementing the buffer.

ajplotkin:
Numbers below, as promised. Pi 4 (arm64, 2 GB), four Nest cameras, real doorbell presses this evening — I was on site, which is the only reason the starved-ring case was testable at all.

## Your CPU correction is right, and it's bigger than you stated

You're correct that the `-vf` on the mjpeg output forces a decode, and that `copy` therefore saves less on a branch carrying this PR. Measured on the Pi against 20 s of real front_door footage (1600×1200, h264 Main), same fixture for every row, user CPU:

| | | user CPU |
|---|---|---|
| A | `copy` only (my config — snapshots come from go2rtc) | **0.061 s** |
| B | `copy` + this PR's mjpeg snapshot output | **6.974 s** |
| C | transcode only | 81.101 s |
| D | transcode + this PR's mjpeg snapshot output | 81.579 s |
| E | decode alone (`-f null -`) | **6.849 s** |

Two rows show the mechanism rather than just agreeing with you:

- **B ≈ E** (6.974 vs 6.849). Almost all of B's cost is the decode; the mjpeg encode and mux together add ~0.1 s, which is within run-to-run noise. So on your branch `copy` is not "remux only" — it's "decode only". Row A is the control that makes this probative: strip the filtered output and the same command drops to 0.061 s.
- **D ≈ C** (81.579 vs 81.101). Bolting the snapshot output onto a transcode costs ~0.5 s, because that decode was already running for the encoder. Same decode, shared.

So my `remux only` cell in #235 is true only when the recording command carries no filtered output, and I stated it generally. I'll add that qualifier — thank you, that's a real correction.

**On magnitude, the honest version is narrower than a single ratio.** C and D use my own encoder settings (default preset, `-profile:v baseline -level:v 4.0`) — not the plugin's tuned args, and not yours, since your tuning now survives in the RTSP branch. A faster preset shrinks the transcode side and the ratio with it, so I'm not going to quote you a number for a branch I haven't measured. With these settings it's 81.6 s → 7.0 s (~12×).

The figure that *doesn't* move with preset is the floor: **B ≈ E means copy's cost on a snapshot-carrying branch is the decode itself** — ~6.8 s user CPU per 20 s of 1600×1200 on a Pi 4. No encoder choice removes that, and it's the number worth designing against.

## Fragment cadence under `copy`

You wanted this measured before #237/#238 could be called settled, so: **cadence is regular, not erratic.**

Two real doorbell presses, captured end to end:

| | warm ring | starved ring (fallback) |
|---|---|---|
| fragments | 21 over 22 s | 13 over 20 s |
| steady-state gap | **2 s ×9** | **2 s ×9** |
| close reason | `0` NORMAL | `0` NORMAL |

Nine 2 s gaps in each run, on both code paths — 2 s at my log's 1-second resolution, so read that as "regular at ~2 s" rather than as precision I don't have. Losing the `-force_key_frames` guarantee turns out not to matter here because the source GOP is already a steady ~2 s, so fragments land on a metronome well inside the hub's timeout.

The sub-2 s gaps differ by run and neither is steady state: in the warm run they're the buffered history draining from memory over the first several seconds; in the starved run — which had no history at all — they're startup and cadence jitter on the fresh dial.

**Audible playback confirmed** — both clips play with sound in the Home app. `Stream #0:1 -> #0:1 (opus (native) -> aac (libfdk_aac))` with `-profile:a aac_eld`, video `copy`, `closeRecordingStream reason=0`.

## The tier boundary — the case you said you'd most like measured

I forced it — ring configured, no history yet for that camera — and it does something specific:

```
[8/1/2026, 5:07:47 PM] Doorbell ring! Front door Doorbell
[8/1/2026, 5:07:48 PM] Recording request received.
[8/1/2026, 5:07:48 PM] [prebuffer:front_door] no init segment yet; falling back to a direct go2rtc dial
[8/1/2026, 5:08:08 PM] [hksv-diag] closeRecordingStream streamId=1 reason=0 Front door Doorbell
```

The recording succeeds and the hub accepts it. First media out at **T+1 s**, against **T+0 s** on the warm-ring path where history is served from memory.

Your ordering — ring first, prewarm as fallback, cold dial last — stands as you wrote it. The one amendment I'd offer is that the bottom rung only exists if nothing else is already keeping the session warm. Here the fallback is

```
ffmpeg -rtsp_transport tcp -analyzeduration 500000 -probesize 5000000 -i rtsp://127.0.0.1:8554/front_door ...
```

and on *this* box that dials a stream go2rtc is already holding, because I keep a preload consumer on it. So there's no SDM session to establish and the cost `motionPrewarm` exists to hide has already been paid — it would be reclaiming about a second. Without that preload go2rtc dials Nest on demand and the establishment cost comes straight back, so this is a property of my configuration, not of go2rtc.

**The real cost of a starved ring isn't latency.** The warm-ring recording was handed ~8.3 s of buffered history, reaching ~6 s before the latched event timestamp (I request `selected + 2 s` of slack; the hub then keeps the 4 s it selected). The fallback clip began at the trigger — strictly ~1–2 s after it, which only widens the gap. Both closed `NORMAL`, so from the plugin's side both "worked", while the fallback produces precisely the clip shape the People/Animals/Vehicles filter discards, because the subject has already left frame. The fallback saves the recording and loses the reason for it. That's the argument for the ring, and it's also why "did it record?" is the wrong success metric.

Being straight about the limit: I induced that condition by killing the ring reader. I haven't measured how often it arises on its own, so I'm not claiming it's common — only that when it happens, this is what happens.

## On `prebufferLength`

Agreed, and thanks for putting it to @potmat — it matches what I found for #233. The field is mandatory in `CameraRecordingOptions` and its documentation demands at least 4000 ms, though nothing enforces it: `RecordingManagement` writes the value into the TLV unvalidated, and my own build advertises `0` when the feature is switched off. So a plugin following the contract advertises at least 4 s whether or not anything backs it, and removing the field isn't available. Implementing the buffer is the only thing that makes the advertisement true.

## Two smaller things

The anchoring detail you flagged is worth restating because it took me a while to get right: history has to be requested relative to the **latched event timestamp**, not `Date.now()`. By the time HomeKit asks, several seconds have passed, and anchoring to "now" quietly discards exactly the pre-trigger window the ring exists for. I also clamp that anchor to 30 s, because the latched timestamp isn't cleared and isn't set for chime-triggered recordings, so it can be hours stale and would otherwise pass the entire ring through the filter.

And on sizing from the hub's **selected** `prebufferLength` rather than the advertised maximum — the hub appears to trim to its selection regardless, so anything beyond it is muxed and then thrown away.

## Offer stands

Happy to run more against these four cameras. If there's a specific case you want measured, say what and I'll capture it the same way — it's just log capture around a real press, so it's easy to repeat.

littlepope81:
Two corrections to my last comment, then the cadence numbers.

**I said HomeKit hadn't requested a recording. Wrong — I grepped my own log for the wrong string.** It had. Six recordings since, on two WebRTC cameras. One 21s clip, `moof` boxes at:

```
12:50:09  :11  :13  :15  :17  :19  :21  :23  :25  :27  :30
```

11 fragments, ~2s apart, 3s max gap, against the negotiated 4000ms. Six requests, zero overlap rejections. Audio input `opus, 48000 Hz, stereo` into the AAC-ELD encode. Your reasoning holds on a stock WebRTC input: source GOP is already ~2s, so losing `-force_key_frames` doesn't matter.

One thing on #238: every recording logs `[mp4] Timestamps are unset in a packet for stream 0` once, at the first fragment. Non-fatal, cadence stays regular — but it's the monotonic-DTS risk you flagged, showing up.

**I also told @potmat `prebufferLength` can't be removed because of an enforced 4000ms floor. You're right that it isn't enforced** — the field is mandatory, the floor is documentation only, and your build advertising `0` proves it. Withdrawn.

**Your #216 is not a nice-to-have.** My subscription had been dead 13.5 hours and I hadn't noticed: zero `Event received` overnight, then all 2547 events arriving at a container restart, ~1170 of them flushed and correctly binned as stale in a three-second window. Your PR description reproduced exactly. Cherry-picked; four days since without a manual restart.

**On the starved-ring fallback** — "saves the recording and loses the reason for it" is a better framing than mine, and it weakens the case for prewarm-as-tier-2 more than I allowed. Conceded.

And your CPU measurement is the right way to state my correction: B≈E means copy is decode-only on a snapshot-carrying branch, not remux-only. The floor is the decode.

ajplotkin:
Thank you for the corrections, and for the cadence numbers — `12:50:09 :11 :13 :15 :17 :19 :21 :23 :25 :27 :30`, 11 fragments, 3s max gap against a negotiated 4000ms, on a **stock** WebRTC input. That's the measurement I couldn't make. Mine came off a prebuffer ring; yours off the plain path, and they agree. The source GOP is ~2s and `-force_key_frames` was never doing anything for this input.

Your `#216` report is the most useful thing anyone has told me about that PR. A subscription dead **13.5 hours** with no symptom until a restart flushed 2547 events, ~1170 correctly binned as stale — that is exactly the failure it was written for, and I had no independent confirmation until now.

## The `Timestamps are unset in a packet` line you flagged

You called it "the monotonic-DTS risk you flagged, showing up." I see the same line, once per recording, on my prebuffer path. I now think it's benign — and I can say why, because the non-benign version of it cost me a doorbell three days ago.

I have written it up upstream as [AlexxIT/go2rtc#2404](https://github.com/AlexxIT/go2rtc/issues/2404), so the detail is there rather than buried in this thread. Short version: go2rtc's Nest producer dies and is re-dialled (`[streams] retry=0`). `internal/streams` then moves existing consumers onto the new connection's tracks via `receiver.Replace()`, so a **fresh RTP stream — new sequence numbers, new and unrelated timestamp bases — is spliced into a consumer's still-open RTSP session.** Nothing structured signals this: go2rtc's RTSP leg sends no RTCP SR, the session is never renegotiated, and with Nest the SSRC doesn't even change (it's 7777 on every camera). The only in-band evidence is a bare sequence jump, which ffmpeg logs as `RTP: PT=..: bad cseq` and then ignores, continuing on its old timestamp base.

When the new base lands *behind* the old one, the mux path clamps every subsequent DTS of that track to `prev+1`. On 2026-08-03 a re-base landed hours behind, and my ring's **audio** timeline froze — structurally valid fragments carrying a dead clock, for **3h53m**, with video completely unaffected (`-codec:v copy` passes timestamps through, and that track's re-base happened to land ahead). It was invisible until the first doorbell press needed the ring, at which point the recording died in 1 fragment and the hub closed it `reason=5` — twice, because HomeKit retried. 611 and 657 `Queue input is backward in time` lines in those two clips; note those come from libfdk_aac's *encoder frame queue*, not the muxer, which is why they're direct evidence the decoder input wasn't advancing.

**Why this matters for your decision rather than mine:** a short-lived consumer can't accumulate this. Each fresh dial re-anchors on the current base, and one that happens to span a splice fails loudly and recovers on retry. Stock dials per recording. Your `motionPrewarm` spawns a reader per event. **Both are immune by construction.** It's a cost of the always-on ring specifically — the thing you said "open yours" about — so you should have it before deciding whether that trade is worth it.

Your single `Timestamps are unset` at the first fragment is the harmless relative: one packet without a timestamp at stream start, not a re-based track mid-session.

## An own goal worth reporting

Chasing this, I found my go2rtc fork's stall watchdog was closing a Nest session after **4s without a keyframe**. Over 48h that fired 23 times, and 13 of those killed a session whose video had arrived **under 2s earlier** (median 1.28s, several at 40–120ms). Healthy Nest video jitters keyframes into the 3–4s range routinely, so the deadline sat inside normal behaviour — and every kill forced a re-dial, i.e. another splice. I was manufacturing the hazard I was investigating.

Raised it to 60s a day ago. Since: **0 keyframe closes** (was ~10/day), 35 recordings all `reason=0`, zero frozen timelines. The watchdog stays because it's the only thing that can replace a session while video still trickles but keyframes have stopped for good — I have one trace at `idr_age 42.8s` with `video_age 1ms` — but 4s was measuring jitter, not droughts.

Worth flagging for anyone reading this thread who runs a keyframe watchdog: mine also meant every drought measurement I had was **censored at 4s**. I'd been quoting drought behaviour while holding a ruler that stopped short of it.

## Where that leaves the prebuffer

Still worth having — starting a recording after the subject has left is the actual defect, and nothing else addresses it. But the honest pitch now includes: a long-lived ring can be silently poisoned by an upstream re-base, it needs detection (I match `bad cseq` on the reader's stderr, plus a burst threshold on muxer DTS clamps, and kill the reader to force a re-anchor), and the failure is invisible without it. If I open it as a PR it will carry both, and the README will say so.

ajplotkin:
You said "Open yours." It's open: [#240](https://github.com/potmat/homebridge-google-nest-sdm/pull/240).

It carries both detectors, as I said it would — the `bad cseq` rule and the burst-thresholded DTS-clamp backstop — so nobody adopts the ring without the thing that makes a poisoned ring visible.

**But there's a limitation that lands specifically on you, and I'd rather say it here than have you find it in the diff.** The ring reads from a local RTSP restreamer, and it's inert unless you configure one. It can't read SDM directly: the ring has to hold a stream open continuously — that's the entire mechanism, history can't be recovered after the fact — and against SDM that means one permanently-open WebRTC session per camera, renewed every ~5 minutes, forever. A restreamer collapses that to one upstream session.

Your cameras aren't going through go2rtc, as you pointed out. So **#240 does nothing for your setup as it stands.** That isn't a detail I want buried.

Which changes something I said. I argued the starved-ring fallback weakened the case for prewarm-as-tier-2, and you conceded it. I now think my argument was narrower than I made it sound: it holds for someone who already runs a restreamer, and doesn't touch the case where there isn't one. For that user `motionPrewarm` isn't a weaker tier of the same feature — it's the only one of the two that runs at all. It still can't hold a frame from before the event, and that limit is real, but "hides the dial and keyframe time" is worth something on its own and I was measuring it against a baseline you can't reach. I'd withdraw the framing, not the measurement.

What I'd genuinely value from your side, if you're willing: **your config is exactly the fallback path.** With no `prebufferRtspBase` set, `handleRecordingStreamRequest` should take the untouched `getStreamer()` route and your recordings should be bit-for-bit what they are today. That's the claim I can't test — every camera here goes through go2rtc, so I can only verify the fallback synthetically. If you cherry-pick it and your six recordings behave exactly as before, that's the confirmation the PR needs, and it's the one thing I'd ask for over any measurement of the ring itself.

For completeness, on this end: 14.0s of clip out of a 5.1s recording window, ~9.0s recovered, `h264 + opus`, consumer released on destroy and no orphaned ffmpeg. The ring itself has been running here continuously since late July; the 28h window after the detectors went in gave 35 recordings, all `reason=0`, zero freeze trips.

One thing I'll flag as unresolved rather than solved: the ring can't rescue the long tail of Pub/Sub latency. Median delivery here is 7.5s and p75 is 47.8s — a 15s ring cannot span a 48s-late delivery, and for those events the history it serves begins long after the subject has gone. Retention margin isn't latency coverage. Those events stay lost regardless of ring size, and only Google delivering sooner would change it.

Issue #235, opened 2026-07-30
HKSV recording re-encodes video unnecessarily — `-codec:v copy` works, and the `-profile:v` requirement is self-imposed


ajplotkin:
The HKSV recording path re-encodes video with libx264. As far as I can tell it doesn't need to, and the reason it does is a chain of two things that look unrelated:

1. `getController()` advertises `video.parameters.profiles: [H264Profile.HIGH]` for recording — High only, no alternative.
2. `handleRecordingStreamRequest()` then derives `-profile:v` from whatever the controller selected, which can therefore only ever be `high`.

Since Nest emits **Main**, that combination forces a re-encode on every clip. Nothing else requires it.

The `profiles: [2 /* HIGH */]` line appears to be inherited verbatim from hap-nodejs's own `Camera_accessory.ts` example, which is where I'd guess it came from rather than a deliberate choice.

## Why the transcode isn't needed

**Nothing hub-side compares the delivered bytes against the negotiated configuration.** In hap-nodejs, `RecordingManagement._startStreaming` treats each yielded fragment as an opaque buffer — it chunks it for the datastream and never parses `moof`/`mdat`/SPS. There is no profile check, no resolution check, no frame-rate check.

**The strongest evidence is already in this repo: the recording path has never applied a scale filter.** There is no `-vf scale` anywhere in `handleRecordingStreamRequest`, so the plugin has been sending HomeKit whatever resolution the camera emits — not the negotiated one — for years, without hub complaints. If the hub validated delivered media against the selected configuration, that would have failed long ago.

**Precedent:** [scrypted](https://github.com/koush/scrypted) defaults to `-vcodec copy` for HKSV recording (`plugins/homekit/src/types/camera.ts` advertises BASELINE/MAIN/HIGH; transcoding is the non-default branch), at a large install base.

## What changes if you copy

On a Raspberry Pi 4 with a 1600×1200 doorbell, measured:

| | transcode | copy |
|---|---|---|
| CPU | ~1.8 of 4 cores | remux only\* |
| Encoder throughput | ~1.37× realtime | n/a |
| Frame rate | had to be capped to keep up | the source's native rate |
| Video | a re-encode of Google's encode | Google's original bytes |

[...]

Issue #232, opened 2026-07-29
HKSV recordings truncated: recording path uses x264 lookahead/B-frames, which a live source can never recover from


ajplotkin:
### Summary

The HKSV recording path builds its ffmpeg command with x264's defaults — no `-preset`, no `-tune`. That means `preset=medium`, which implies `rc-lookahead=40` and B-frames. Lookahead assumes the next frames already exist: true for a file, false for a live camera. On a Raspberry Pi 4 this was enough to truncate recordings.

Notably the plugin **already gets this right for the live path** — `StreamingDelegate.js` defaults `vEncoder` to `libx264 -preset ultrafast -tune zerolatency`. The recording path just never inherited the same reasoning.

### Why it truncates clips rather than merely slowing them

A live source caps encoder throughput at **1x** — frames only arrive as fast as the camera sends them. So any deficit taken at startup can never be recovered; the clip simply ends early. With `rc-lookahead=40` at 15fps, that's ~2.7s of buffering before the first frame is even emitted.

Observed on a Pi 4 (1600x1200 doorbell, HKSV clip):

```
frame=    0 ... (for 9 seconds)
frame=    1  speed=0.0195x
...
HksvStreamer destroy          <- hub gives up at 18s
[libx264] consecutive B-frames:  1.3%  1.3%  1.9%  95.6%
```

Two fragments instead of eight. The clip was too short to be useful — a package delivery that never showed up properly in Home.

There's a second, independent waste: `-r` is set from `cameraRecordingConfiguration.videoCodec.resolution[2]`, which HomeKit typically negotiates at 30. With a 15fps source ffmpeg duplicates every frame (`dup=38` in the logs) — double the encoder work for no extra detail.

### Fix

Three changes to the recording `videoArgs`:

```js
"-preset", "veryfast",
"-tune",   "zerolatency",
// and cap the output rate at the source rate:
"-r", Math.min(this.cameraRecordingConfiguration.videoCodec.resolution[2], SOURCE_FPS).toString(),
```

Measured on the same camera, same resolution, no other changes:

| | before | after |
|---|---|---|
| fragments | 2 | **8** |
| frames | 159 | **391** |
| clip | cut at 18s | **26.0s, 6.8 MB** |
| B-frame runs | 95.6% in 4-long runs | none |

[...]

PR #238, opened 2026-07-30
Copy the camera H.264 for HKSV recording on the WebRTC path (drop the libx264 transcode)

ajplotkin:
Implements #235, which has the full rationale, the measurements and the caveats.
## Why the transcode isn't needed

Nothing hub-side validates delivered media against the negotiation. hap-nodejs's `RecordingManagement` chunks each yielded fragment and ships it, never parsing `moof`/`mdat`/SPS. The strongest evidence is in this repo: **the recording path has never applied a scale filter**, so the plugin has been sending HomeKit un-negotiated resolutions for years with no complaints. [scrypted](https://github.com/koush/scrypted) defaults to `-vcodec copy` for HKSV at a large install base (transcoding is its non-default branch).

## Verified on the stock WebRTC input

I record from a local go2rtc RTSP stream, so I did **not** want to submit this on the strength of my own setup. I reproduced the stock path exactly: real Nest H.264 + Opus, delivered as RTP over UDP with the SDP on stdin, using `WebRtcNestStreamer`'s own argument string (`-protocol_whitelist pipe,crypto,udp,rtp,fd -analyzeduration 15000000 -probesize 100000000 -fpsprobesize 0 -i -`).

```
output          : h264 (Main), 1600x1200, 24.97s from a 25s input
first frame     : type:I
fragments       : 20 moof over ~25s  =>  1.25s each (negotiated max is 4000ms)
decode errors   : 1
decode errors in the SOURCE capture, never transported : 1  (identical, pre-existing)
frames          : 594 vs 595 in source (one lost to UDP in the harness)
DTS/RTP warnings: none
```

The decode error is in the source capture itself and is not introduced by copy or by RTP transport. That was the specific risk worth checking — RTP jitter feeding the mp4 muxer's monotonic-DTS requirement with no encoder to absorb it.

## Adaptive-resolution cameras: copy cannot be made safe

@littlepope81 found a camera class where this does not work, and it belongs in the description rather than buried in a comment. Newer Nest hardware renegotiates resolution **mid-session**: on one of his six cameras, 640x368 and 1920x1088 inside the same sessions, with 31 and 9 decoder reinits respectively over a day. His other five hold one geometry and never do it. An fMP4 track declares its dimensions once, in the init segment, and cannot re-declare them, so after the first switch the track stops matching its samples: video freezes ~2s in and never recovers while audio plays on. Every clip from that camera.

This section previously argued that master was already broken for these cameras, resting on an A/B whose middle row read "transcode, no scale filter: still freezes". @littlepope81 has since retracted that row, having found the test was confounded by a restart button that silently did nothing. The corrected behaviour:

| | result |
|---|---|
| `-c:v copy` | freezes |
| transcode, no scale filter | plays through |
| transcode + scale pinned to the negotiated resolution | plays through, but squeezes 4:3 cameras |

ffmpeg's `-autoscale` is on by default and locks the encoder's output to the first frame's resolution, so a transcode is geometry-stable with no `-vf` at all. Reproduced independently on ffmpeg 9.0.1 by concatenating a 640x360 and a 1920x1080 elementary stream: under `-c:v copy` the track header declares 640x360 while the file carries 60 frames at each geometry; transcoded with no filter, all 120 frames come out 640x360.

That makes copy a regression introduced by this PR rather than a pre-existing defect it merely fails to fix. On master these cameras record correctly, because the transcode normalises them. Under copy the frames reach the muxer at whatever geometry they arrive with, the fMP4 track cannot re-declare its dimensions, and the clip freezes about 2s in with audio playing on.

The third row is not the fix either: pinning to the negotiated resolution squeezes a 1600x1200 doorbell to 1440x1080 with bars, and nothing hub-side enforces that geometry anyway.

I do not think copy can be made safe for this hardware without detecting the switch and falling back mid-recording, and the switch is invisible on the copy path: ffmpeg never decodes, so no `Reinit context` line prints at any loglevel. Detection would have to parse the SPS.

That leaves the scope of this PR narrower than claimed: a real CPU saving on fixed-geometry cameras, which is still every camera I own and five of @littlepope81's six, and a correctness regression on adaptive ones. Gating copy per camera, or behind an opt-in with this caveat documented, is the honest shape. I would rather land it that way than leave the description arguing something its author knows is false.

Five of his six cameras hold a single resolution and are unaffected, as are all four of mine — so this is hardware-generation-specific rather than universal. But it looks like standard behaviour on newer hardware rather than a fault, so more people will meet it as they upgrade.

## RTSP cameras keep the transcode

`WebRtcNestStreamer` runs a FIR/PLI keyframe-request loop that holds the IDR interval near 2s. `RtspNestStreamer` has no equivalent, and I have not measured Nest's RTSP IDR cadence. If it exceeded the negotiated `fragmentLength`, copied fragments would breach the one limit `-force_key_frames` was guaranteeing — so that branch is gated to keep the encoder until someone can measure it.

## What changes

| | transcode | copy |
|---|---|---|
| CPU | ~1.8 of 4 cores on a Pi 4 (1600×1200) | remux only |
| Encoder throughput | ~1.37× realtime | n/a |
| Video | a re-encode of the camera's encode | the camera's original bytes |

That throughput figure is the margin against the hub's fragment timeout; under concurrent load a transcode can drop below realtime, which is the truncation mechanism in #232. With no encoder there is nothing to fall behind.

## Notes

- **Audio untouched.** `-an` stays exactly where it is; #237 owns audio behaviour. If #237 lands first this needs no rebase — the two edit adjacent but distinct parts of `videoArgs`/`audioArgs`.
- The advertisement is deliberately left as `profiles: [HIGH]`. Changing it alters the supported-configuration hash, which makes hap-nodejs discard every hub's persisted selected configuration and forces a renegotiation on upgrade. Copy needs no such change.
- `tsc --noEmit` clean.
- Running on my own deployment since yesterday: clips play with audio, close `reason=NORMAL`, fragments at ~1.67–2s.
- #227 and #230 also touch this file; happy to rebase around whatever lands first.


littlepope81:
Found a case where copy can't work, on the stock WebRTC path.

One of my cameras is a newer model and re-negotiates resolution mid-stream — 640x368 and 1920x1088 inside the same sessions, 31 and 9 decoder reinits respectively over a day. My other five cameras each hold a single resolution and never do it.

An fMP4 track declares its dimensions once, in the init segment, and can't re-declare them. So after the first switch the track stops matching its samples: video freezes and never recovers, audio plays on untouched. Every clip from that camera, ~2s in. The other five were fine throughout.

A/B on hardware, one variable at a time:

- copy — freezes
- transcode, no scale filter — still freezes
- transcode + scale pinned to the negotiated resolution — plays through

The middle one is the useful result. It's not about the encoder, it's that this path has never applied a scale filter — which you'd already noticed, from the other direction. A fixed-resolution camera never triggers it.

I don't think copy can be made safe for these cameras. It passes frames through at whatever geometry they arrive with, so the change reaches the muxer regardless. Detecting the switch and falling back mid-recording would work but that's real work.

Gone to always-transcode on my branch for now. Worth a caveat in #238 either way, since adaptive resolution looks like standard behaviour on the newer hardware rather than a fault — more people will hit it as they upgrade.

Separately: you were right that the `Timestamps are unset` line is benign. I moved `-fflags +genpts` ahead of `-i`, where it actually belongs, and it changed nothing about the freeze.

ajplotkin:
That's a real limitation and the A/B settles it. Thank you for running it on hardware — I can't reproduce it here, so this is a case I would have shipped blind.

**The middle result is the important one, and I'd put it more strongly than you did.** `transcode, no scale filter — still freezes` means **master is already broken for these cameras**, today, without #238. The recording path has never applied a scale filter, and an fMP4 track declares its dimensions once either way. So this isn't a regression #238 introduces — it's a pre-existing defect that #238 fails to fix and can't fix, because `copy` has no place to put a scale filter.

That changes where the caveat belongs. It isn't "don't use copy, transcode is safe"; it's "this plugin cannot record an adaptive-resolution camera at all right now." The fix — pinning `-vf scale=` to the negotiated resolution — is worth its own PR and helps everyone independently of copy-vs-transcode. I'd rather that landed than have #238 quietly imply transcoding is the safe option, because your own data says it isn't.

## What I see here, and how much it's worth

Four cameras, all older hardware, and I can't reproduce the switch:

- Two resolutions do appear in my logs — `1600x1200` and `1920x1080` — but they're **different paths, not a mid-session change**: 4:3 `yuvj420p` is the doorbell's still-image/snapshot decode, 16:9 is the video stream. No single session shows two video geometries.
- The only resolution changes I've ever induced came from **changing the optimization setting in the Home app**. That renegotiates between sessions, so the next session gets a fresh init segment and nothing breaks. Categorically different from a camera re-declaring geometry mid-session.

The strongest evidence I have isn't the log window, which is short. It's that I run a **continuous `-c copy` fMP4 ring** against these cameras — a long-lived reader that has been up for weeks. That is about the most sensitive possible detector for what you found: a mid-session geometry change would freeze the ring's video permanently, not just for one clip, and it would have been unmissable. It hasn't happened once.

So: absence of evidence on four older cameras, from someone whose setup would have screamed if it were happening. Not evidence of absence on yours, and your point that this is standard behaviour on newer hardware rather than a fault is the part I'd take seriously.

## Disclosure, since it's my problem too

The same constraint applies to the pre-trigger buffer I opened as #240. The ring is `-c copy` into fMP4, so a camera that re-declares geometry mid-session would poison the ring exactly as you describe — and worse than a single clip, because the ring is long-lived: it would keep serving a broken timeline until something restarted the reader.

There's a cheap mitigation that fits the existing machinery. The ring already watches ffmpeg's stderr and kills the reader on an RTP `bad cseq` and on a burst of muxer DTS clamps, both to force a re-anchor. `Reinit context to WxH` with a **different** WxH is the same shape of signal. That doesn't rescue the clip in flight, but it stops one geometry change from silently breaking every subsequent recording. I'll add it if #240 goes anywhere, and it's cheap enough that I'd add it regardless.

If you're willing, the single most useful thing you could tell me is the model name of the camera that does this. If it correlates with a generation rather than a setting, that's the caveat text — and it tells me whether the ring needs the guard before anyone else adopts it or after.

And noted on `-fflags +genpts` — thanks for going back and checking that rather than leaving it as a maybe.

ajplotkin:
Following up on the scale-pinning PR I said was worth writing. Three things
changed my mind about what it should do, and one of them means I can't write it
yet.

Pinning to the negotiated resolution is the wrong target. I had assumed the
output had to match `cameraRecordingConfiguration.videoCodec.resolution`. It
doesn't: nothing hub-side compares delivered bytes to the negotiation. hap-nodejs
treats fragments as opaque and never parses moof/mdat/SPS. That is the same
property #238 relies on — the copy path sends the camera's native Main-profile
stream while the negotiation says High, and it plays.

So a filter pinned to the negotiated frame would downscale for a constraint that
isn't enforced. The advertised list mixes 16:9 and 4:3 (it includes
`[1600, 1200, 30]`), so on a 4:3 camera negotiated to 1920x1080 every clip would
become 1440x1080 with black side bars and lose 120 lines, in exchange for
nothing. The live path already scales with
`force_original_aspect_ratio=decrease` and never pads; recording would have
become the only padded output the plugin produces.

The requirement is constant geometry, not negotiated geometry. Those are
different, and I had conflated them. A fix should lock the output to whatever the
first frame is, which costs no resolution and needs no knowledge of the source.

I can't reproduce the failure, and my attempt to fake it disproved my own test
rather than the fix. I built a resolution-switching input by concatenating
1920x1088 and 640x368 segments. With no filter at all — master's behaviour —
ffmpeg auto-scaled the 640x368 frames up to the first configuration and emitted
40 frames at a constant 1920x1088. No geometry change reached the track, so that
input doesn't exercise the bug, and any "fix" I measured against it would have
been measuring nothing.

That likely means the real path differs from a naive concat in a way that
matters. Two things would let me build a real reproduction:

1. The camera model, and whether the switch shows up as `-1:-1 -> WxH` reinit
   lines or as ffmpeg logging a changed input stream.
2. A short capture of the input as the recording path receives it
   (`-c copy -f mpegts` off the same source across a switch) — even 30 seconds
   spanning one transition. That is the artefact I'm missing.

If the freeze is real on your hardware and ffmpeg is auto-scaling as it did for
me, then the freeze has a different cause than the declared dimensions, and the
scale filter would not have fixed it. I would rather find that out before sending
a PR than after.

I've kept the branch locally. It is one `-vf` on the recording args, and I'll
finish it against a real capture rather than a synthetic one.

littlepope81:
You were right, and my reason for keeping the filter was wrong.

ffmpeg's `-autoscale` is on by default and locks output to the first frame's
resolution, so a transcode is already geometry-stable with no `-vf` at all. On
ffmpeg 8.0, which is what my bridge runs:

```
-c:v copy                     both geometries end up in the track
-c:v libx264, no -vf          constant throughout
```

`-noautoscale` breaks it ("Input picture width (1920) is greater than stride
(640)"), but nothing on this path passes that. So the filter was never what
stopped the freeze. It just picked a different constant size, and since it
pinned to the negotiated resolution it was squeezing my 1600x1200 doorbell to
1440x1080 with bars. I've removed it.

I also need to correct my commit message on that change. It said forcing the
transcode alone did not fix the freeze. That isn't true — the A/B behind it was
confounded by a restart button that silently did nothing.

On your concat test: it ran through a transcode, so autoscale normalised the
switch before it could reach the muxer. That path can't show the bug whatever
the input does. Worth retrying with `-c copy`.

Here's the capture you asked for. The camera does still switch, mid-recording:

```
Recording input geometry 640x360.                        Backyard camera
Recording input geometry changed 640x360 -> 1920x1080.   Backyard camera
```

It shows up as decoder `Reinit context to WxH` lines, and those only print at
`-loglevel verbose`. The plugin passes no `-loglevel`, so it runs at info and
they never appear, which is why there was nothing in the logs for either of us.
Two things that cost me time if you go digging: `Reinit` reports the coded
16-aligned size, so a 1920x1080 stream shows as 1920x1088, and the filtergraph
input line that carries the real size is formatted differently on ffmpeg 6 and 8.

So copy isn't safe on this camera, and I don't see a way around that short of
detecting the change and falling back mid-recording.

One thing you might care about: first-frame locking means a clip that starts
while the camera is degraded stays that way for its whole length. Mine started
at 640x360, rose to 1920x1080, and the whole clip stayed 640x360. I'm now
learning each camera's largest seen geometry and pinning to that, which fixes
it, but it means upscaling during the degraded stretches — roughly 10x the
encode CPU in a rough benchmark. Free on my x86 box, probably not on your Pi 4,
so I'd want your numbers before arguing it belongs upstream.

ajplotkin:
Retried with `-c copy`. You were right that my concat test could not show anything: it
transcoded, so autoscale had normalised the switch before the muxer saw it.

Redone on ffmpeg 9.0.1, concatenating a 640x360 and a 1920x1080 H264 elementary stream
(30 fps, 2s each) and muxing to fragmented MP4:

| | track header declares | frames in the file |
|---|---|---|
| `-c:v copy` | 640x360 | 60 at 640x360, then 60 at 1920x1080 |
| `-c:v libx264`, no `-vf` | 640x360 | 120, all 640x360 |

Same as what you saw on 8.0.

One thing from the same run that bears on detect-and-fall-back: under `-c copy` ffmpeg never
decodes, so the switch produces no `Reinit context` line at any loglevel. I get one at verbose
and debug, the initial probe, and none at info or warning. Decoding the same stream gives
three. So on the copy path nothing on stderr marks the change, and detection has to read the
bytes: `-bsf:v trace_headers` at trace level shows the second SPS, or parse
`pic_width_in_mbs_minus1` yourself.

That also retires something I claimed in August, that my prebuffer ring was a sensitive
detector for this. It runs `-c copy` at a low loglevel, so it could never have seen a switch.
My only real check was a 90-second geometry sample, which against 40 reinits a day would have
caught one about 4% of the time. So I have had no evidence either way about my own cameras.
Since this afternoon I am sampling geometry every 15s for 24h off the snapshot JPEGs a warmer
already writes, which costs the cameras nothing. It will catch a sustained switch, not a brief
one. Zero changes in the first hour, and my doorbell is 1600x1200, the same 4:3 as yours. I
will post the result either way.

With the middle row of that A/B gone, the description on #238 is wrong where it argues master
is already broken for this hardware. Master transcodes, autoscale holds it to the first frame,
and copy is what breaks. That makes copy a regression for adaptive-resolution cameras rather
than a pre-existing defect it merely fails to fix, which is a different and worse position for
the PR to be in. I am rewriting that section and the table now, and correcting "two of his
cameras" to one.

I have not measured the upscale cost on the Pi yet and will post numbers.

ajplotkin:
Pi numbers, as promised, with the caveats first because they matter for how you read them.

Raspberry Pi 4 Model B, 4 cores, 2GB, distro ffmpeg 7.1.5 rather than the bundled 6.0 static
build. Synthetic `testsrc2` input, which is cheap to decode. 300 frames, so 10s of video per
run. `libx264 -preset ultrafast -tune zerolatency`: that is what my build passes on the
recording path, but master passes no preset there (`StreamingDelegate.ts` line 497 sets
`-codec:v libx264` and nothing else), so master encodes at x264's default and every figure
below is a floor for it, by a lot. Each encode ran in its own cgroup with a 450M cap and swap
denied, niced under the camera stack, which stayed live throughout.

Single stream, all four capped to two cores so they compare like for like:

| | realtime | peak RSS |
|---|---|---|
| remux copy | 24x | 53 MB |
| transcode 640x360, no scale | 5.4x | 78 MB |
| upscale 640x360 to 1920x1080 | 1.1x | 130 MB |
| native 1920x1080 | 1.2x | 169 MB |

Concurrent upscales, no CPU cap:

| streams | realtime |
|---|---|
| 1 | 1.4x |
| 2 | 0.8x |
| 4 | 0.4x |

The useful result is in the first table rather than the second: upscaled 360p and native 1080p
cost the same within noise, 1.1x against 1.2x and 130MB against 169MB. The scale step is
nearly free and the 1080p encode is the whole cost, so pinning to the largest seen geometry
costs what a healthy camera at that geometry already costs. Your 10x is against a 360p encode,
which is the same observation from the other end.

What this board cannot afford is a second concurrent 1080p encode, pinned or not. `-tune
zerolatency` sets `sliced_threads`, and x264 then takes one thread per core, so a single
stream already spans all four. The second table is one box's throughput divided N ways:
multiply it out and it is flat at about 1.5 streams of 1080p work in total. So a 1080p
transcode fits for one camera at a time and not for two. Free on your x86 box, as you
expected, and not something I can carry.

Your approach holds up; the numbers argue only against it being the default. As an opt-in for
people with an adaptive camera it is clearly right, since the alternative is clips that freeze.
A per-camera setting would keep it off the Pis.

The same box says something about the copy path that I had underweighted, and I had it badly
wrong at first: my initial RSS figures summed every process named ffmpeg, which on this box
includes the prebuffer rings and the snapshot warmer, and were inflated by about 48MB each.
Measured per process, the four continuous `-c copy` rings running here right now are about
12MB and 1 to 2% of a core each, so under 50MB and a tenth of a core for all four. Four
continuous native 1080p transcodes would want roughly 1.7 cores each at ultrafast, about 7
cores on a 4-core box, so they cannot run at all regardless of memory. That cuts both ways for
#238: the regression on adaptive cameras is real and I have corrected the description, but on
constrained hardware copy is what makes a continuous per-camera reader possible.

Which points at gating rather than choosing: copy by default, transcode-and-pin for cameras
known to renegotiate. The open problem is still detection, and on the copy path there is
nothing in the logs to detect with.

